PR #3, opened 2021-06-11
Cargo Config Update

FloppyDisck:
Original cargo config was trying to work with the rust projets in the contracts folder. This was changes to only point to the shade-contracts folder.


mohammedpatla (approved):
Looks good to me

PR #15, opened 2021-07-25
Improved Scripted Contract Testing

FloppyDisck:
_Overview_
* Created base functions for interacting with the secretcli
* Created a base contract class to store all contract values
* Added a Snip20 implementation of said class with basic support for our tests (will probably be expanded at a later time)
* Added a Mint implementation along with all of its Handles and Queries
_Details_
The new testing script contains functions that automate using the secretcli. Using these functions is a new base contract class that can be expanded with the necessary functionality to abstract users writing tests as much as possible.

### DIFF
--- a/contracts/compiled/contract_tester.py
+++ b/contracts/compiled/contract_tester.py
@@ -1,128 +1,42 @@
-import subprocess
-import json
-import time
+import string
+import random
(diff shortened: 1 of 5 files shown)


mohammedpatla (approved):
Looks good to me mostly for testing!!

PR #18, opened 2021-07-28
Added mint contract migration

FloppyDisck:
_Overview_
* Added mint migration
* Extended python contract testing tool-set

_Details_
Minting contract now contains a migration function. this will deactivate the contract and move all of its private state information into the new one.


contracts/compiled/contractlib/contractlib.py
```diff
@@ -2,20 +2,24 @@
 
 
 class Contract:
-    def __init__(self, contract, initMsg, label, admin='a', uploader='a', gas='10000000', backend='test'):
+    def __init__(self, contract, initMsg, label, admin='a', uploader='a', gas='10000000', backend='test', wait=6):
```
mohammedpatla, line 5: why wait=6?

mohammedpatla (approved):
Approving

contracts/compiled/contractlib/contractlib.py
FloppyDisck: Any command that modifies the ledger is usually met with some delay in receiving info resulting in the pipeline returning nil values. A quick and dirty way to avoid this is just setting up a delay.

Im looking into how to fix this

mohammedpatla: Ok, looked weird hence i caught it. Hope you are noting these small changes to be made in the future code. 

For mvp no one cares, but we dont want to leave vulnerability for final code.

FloppyDisck: This is also the tester, it doesnt affect the contract

PR #31, opened 2021-08-04
oracle should use USD not tether USDT

DrPresident:


mohammedpatla (approved):
Required change, not to rely on USDT prices

PR #35, opened 2021-08-06
Mint/minimum expected mint value

### LINKED ISSUE #12: Mint contract - Price Slippage Revert Contract
Front end app inputs a minimum quantity of Silk the user should receive in the transaction (via user entered slippage % tolerance) based on calculations done on the front end from an Oracle price query. If the Silk qty received drops below the minimum by the time the transaction is processed, then the transaction is cancelled, no SILK is received and no sSCRT is burned.

FloppyDisck:
_Overview_
* Mint contract now supports msg arguments when receiving from snip20

_Details_
In the future, users will be sending different assets with the intent to use various minting procedures. To achieve this the contract must be able to handle the msg argument properly. The currently supported structure makes way for possible expansion with little to no effort.

_Issues_
Closes #12


mohammedpatla (approved):
appriving @DrPresident you should review this.

PR #52, opened 2021-08-17
Mint | Experimental Restructure

FloppyDisck:
_Overview_
* Sending to mint is more flexible
* Removed redundant commands
* Stored asset queries can now be modified
* Contract tester is now cleaned up

_Details_
Mint contract now supports a more flexible minting/burning algorithm and just required the target coin. This removes alot of the complexity of the contract in favor of something easier to read, understand and use. Because of this, the shade and silk contract do not need to be explicit, they are just part of the list of mintable assets (can now be expanded without changing the code). Apart from thtis, some redundant code in the contract tester was removed in favor of a more readable and cleaner solution.

_Missing_
Public testnet tester implementation


packages/shade_protocol/src/mint.rs
```diff
@@ -42,16 +40,13 @@ pub enum HandleMsg {
     },
     UpdateConfig {
         owner: Option<HumanAddr>,
-        silk: Option<Contract>,
-        shade: Option<Contract>,
         oracle: Option<Contract>,
     },
     RegisterAsset {
+        name: Option<String>,
```
DrPresident, line 46: Can we query this from the snip-20? Same for `SupportedAsset`

packages/shade_protocol/src/mint.rs
```diff
@@ -42,16 +40,13 @@ pub enum HandleMsg {
...
         contract: Contract,
-    },
-    UpdateAsset {
-        asset: HumanAddr,
-        contract: Contract,
+        burnable: Option<bool>,
```
DrPresident, line 48: What do we do if it isn't burnable? This feels like we're setting a value but I don't believe we can in this context

packages/shade_protocol/src/mint.rs
```diff
@@ -8,26 +8,24 @@ use crate::msg_traits::{Init, Handle, Query};
...
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct InitMsg {
     pub admin: Option<HumanAddr>,
-    pub silk: Contract,
-    pub shade: Contract,
     pub oracle: Contract,
-    pub initial_assets: Option<Vec<AssetMsg>>,
+    pub initial_assets: Option<Vec<SupportedAsset>>,
```
DrPresident, line 28: I tried using this parameter in my testing and I really like the idea, but once I tried to do like 8 or more assets I would run out of gas, and as far as I could tell the instantiate command only lets you use 200,000 gas for some reason. With a limit that tight I think we should omit this from the instantiate parameters and add a `RegisterAssetBulk` message where we can at least adjust our gas based on how many assets we want to register.

contracts/mint/src/state.rs
```diff
@@ -25,10 +25,10 @@ pub fn asset_list_read<S: Storage>(storage: &S) -> ReadonlySingleton<S, Vec<Stri
     singleton_read(storage, ASSET_LIST_KEY)
 }
 
-pub fn assets_r<S: Storage>(storage: & S) -> ReadonlyBucket<S, BurnableAsset> {
+pub fn assets_r<S: Storage>(storage: & S) -> ReadonlyBucket<S, SupportedAsset> {
```
DrPresident, line 28: Note: I like the `*_r/w` format a lot, we should make sure to use it all our states

DrPresident (approved):

contracts/mint/src/state.rs
FloppyDisck: Agree, we could make another PR for this to resolve on all of the contracts.

packages/shade_protocol/src/mint.rs
FloppyDisck: Agree, we could add a command to add multiple assets. Also I think instantiation supports setting a gas amount, if not, the parameter is more geared towards migration so it shouldnt be an issue

packages/shade_protocol/src/mint.rs
FloppyDisck: The term burnable is more about the snip20 being burnable, so for example sSCRT is not burnable, so instead of actually calling the snip20's burn interface we just store the amount in the mint contract technically being burnt.

packages/shade_protocol/src/mint.rs
FloppyDisck: Yes, when querying a specific asset all of these values will be returned

packages/shade_protocol/src/mint.rs
DrPresident: You maybe can, but I had my gas set for everything at 10000000 and it worked for everything but I couldn't find a way to change the gas for instantiate. Feel free to double-check that and correct me.

PR #69, opened 2021-08-27
Mint | add mint support to new mint contract

FloppyDisck:
_Overview_
* New mint contract can now add assets
* Contract can now mint and burn accordingly
* Testers now sample this contract
* Replaced protocol msg traits with the ones included in secret_toolkit

_Details_
The micro_mint contract has been updated to a deploy ready state for a future replacement of the existing mint contract.


packages/shade_protocol/src/micro_mint/mod.rs
```diff
@@ -48,6 +48,7 @@ pub enum HandleMsg {
     },
     RegisterAsset {
         contract: Contract,
+        burnable: bool,
```
DrPresident, line 51: When we can we should get this with the "token_config" and omit it from the parameters

packages/shade_protocol/src/snip20.rs
```diff
@@ -8,4 +8,5 @@ use secret_toolkit::snip20::TokenInfo;
 pub struct Snip20Asset {
     pub contract: Contract,
     pub token_info: TokenInfo,
+    pub burnable: Option<bool>,
```
DrPresident, line 11: Once we get this queried from the snip20, we should add the other values as well (that don't require extra queries) and make this not optional

DrPresident (approved):

packages/shade_protocol/src/micro_mint/mod.rs
FloppyDisck: Agree, this change is now included in 6c71e

packages/shade_protocol/src/snip20.rs
FloppyDisck: Resolved in 6c71e

PR #75, opened 2021-09-02
Oracle sswap

DrPresident:
Oracle sswap integration

`register_sswap_pair`, takes in an `Contract` referencing an existing `secreteswap_pair`, this pair must be an sSCRT/*. The non-sscrt token will be "registered" e.g. registering a pair for sSCRT/SETH will cause `get_price(SETH)` to query the price from SecretSwap. 

`get_price` for an sswap asset will get the price of SCRT in USD, and use that to translate the trade amount received from 1sSCRT for the asset into the asset/USD price

NOTE: The Hard-coded 11.47 SHD price has been removed, queries for this symbol will fail unless we have a registered SecretSwap pair with enough liquidity


FloppyDisck (approved):

FloppyDisck:
Missing details about updating mint contract

PR #76, opened 2021-09-03
Snip initializer

FloppyDisck:
Created a new contract that initialized the Shade/Silk snip20s and stores the initial balances so the community can query them


DrPresident:
This contracts function is already being handled by the deployment python scripts

FloppyDisck:
@DrPresident yes but as per last meetings tasks, @mohammedpatla asked for a contract that deploys both snip20s and records their initial balances for public viewing

mohammedpatla:
Ok as long as this initialized with balance for admin

mohammedpatla (approved):
As long as this is s version works with a drop on admin.

DrPresident (approved):

PR #81, opened 2021-09-04
Micro mint documentation

FloppyDisck:
Updated the documentation for the micro_mint

Also renamed some functions/structs to conform to the rest of the contracts


DrPresident (approved):
The `micro_mint/mod.rs` thing we can move back,  I was just playing around with different structures.

FloppyDisck:
> The `micro_mint/mod.rs` thing we can move back,  I was just playing around with different structures.
> 
> Alternatively we could make them each like this and move more of the functionality into the shade_protocol package but I'm fine with either

I think we should stick with the original formatting because that structure adds more clutter when developing

PR #87, opened 2021-09-10
Contract tester in rust

FloppyDisck:
Added a new tester that runs completely on rust that utilizes the defined types in the shade_protocol package to run everything


makefile
```diff
@@ -2,19 +2,24 @@ contracts_dir=contracts
 compiled_dir=compiled
 checksum_dir=${compiled_dir}/checksum
 
-CONTRACTS = mint snip20 treasury micro_mint oracle mock_band initializer
+define prepare_wasm =
```
DrPresident, line 5: Could this be `build_wasm`?

packages/shade_protocol/src/band.rs
```diff
@@ -1,7 +1,17 @@
...
+use secretcli::secretcli::{TestInit, TestHandle, TestQuery};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct InitMsg { }
+
+impl InitCallback for InitMsg {
+    const BLOCK_SIZE: usize = 256;
+}
+
+impl TestInit for InitMsg {}
```
DrPresident, line 14: I've seen something like `#if(cfig)` to only do these when for tests, I'll try to find exactly what it is

DrPresident (approved):

makefile
FloppyDisck: yeah, ill rename it now

PR #88, opened 2021-09-12
minting cap

FloppyDisck:
Minting now has epochs, epochs contain a frequency in which each epoch lasts and a cap on the total mintable of the asset.

Some improvements to the rust testing project were also done, for example colored text for better debugging and adding the epoch tests


contracts/micro_mint/src/contract.rs
```diff
@@ -42,6 +40,24 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
...
+            Some(frequency) => frequency.u128() as u64,
+        },
+        mint_capacity: match msg.epoch_mint_limit {
+            None => Uint128(0),
+            Some(capacity) => capacity
+        },
+        total_minted: Uint128(0),
+        next_epoch: match msg.epoch_frequency {
+            None => 0,
+            Some(frequency) => env.block.time + frequency.u128() as u64,
```
DrPresident, line 56: Is there a way we can set this to a specific time of day UTC?

contracts/micro_mint/src/handle.rs
```diff
@@ -124,13 +117,36 @@ pub fn try_burn<S: Storage, A: Api, Q: Querier>(
...
     if amount_to_mint < msgs.minimum_expected_amount {
         return Err(StdError::generic_err("Mint amount is less than the minimum expected."))
     }
 
+    // Check against mint cap
+    let mut limit_storage = limit_w(&mut deps.storage);
+    let mut limit = limit_storage.load()?;
+
+    // When frequency is 0 it means that mint limits are disabled
+    if limit.frequency != 0 {
```
DrPresident, line 132: I think we want this mint limit to be public, and if so could this be an optional field like "limit" under the config? then we can just omit it in the case we don't have a limit

DrPresident (approved):

contracts/micro_mint/src/handle.rs
FloppyDisck: You can query the limit with GetMintLimit

contracts/micro_mint/src/contract.rs
FloppyDisck: Good idea, when initializing or updating limit give the admin the option to  set the next_epoch too

FloppyDisck: Added an issue for this

PR #97, opened 2021-09-12
Mint | secondary burn

FloppyDisck:
Added the ability to send non burnable tokens to a given throwaray address

### DIFF
--- a/contracts/micro_mint/src/handle.rs
+++ b/contracts/micro_mint/src/handle.rs
@@ -87,20 +87,30 @@ pub fn try_burn<S: Storage, A: Api, Q: Querier>(
         }
     }
 
-    //TODO: if token_config is None, or cant burn, need to trash
-
     // Try to burn
-    match burn_asset.asset.token_config {
-        Some(ref conf) => {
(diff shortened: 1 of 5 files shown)


contracts/micro_mint/src/handle.rs
```diff
@@ -87,20 +87,30 @@ pub fn try_burn<S: Storage, A: Api, Q: Querier>(
...
+            messages.push(burn_msg(burn_amount,
+                                   None,
+                                   256,
+                                   burn_asset.asset.contract.code_hash.clone(),
+                                   burn_asset.asset.contract.address.clone())?);
+        }
+        else {
+            // If no config then dont burn
+            if let Some(recipient) = config.secondary_burn {
+                messages.push(send_msg(recipient, amount, None, None, 1,
```
DrPresident, line 102: This should burn `burn_amount` so it isn't burning the capture

FloppyDisck: fixed in 116f39b

DrPresident (approved):

PR #99, opened 2021-09-18
Shade protocol tool

DrPresident:


packages/shade_protocol/src/micro_mint.rs
```diff
@@ -124,6 +128,7 @@ impl Query for QueryMsg {
     const BLOCK_SIZE: usize = 256;
 }
 
+#[cfg(test)]
```
FloppyDisck, line 131: These should be removed since secretcli doesn't have anything to do with cargo tests
Maybe a specific flag should be created in the future to prevent compiling this library when releasing the module

FloppyDisck (changes requested):

packages/shade_protocol/src/micro_mint.rs
DrPresident: Im pretty sure thats what that flag is, it will only do that in a test context. Is there a reason we would want this implemented outside of a test context?

FloppyDisck: Its better to not apply these changes until another PR, since it would require the on-chain rust tests to be changed

DrPresident: Why? If they are tests, shouldn't it be under cfg(test)?

FloppyDisck: Yes, but it would require editing the library, all of the shade_protocol files and the network_tester folder
So its better to leave this to another PR unless you prefer to do those changes here

FloppyDisck: These flags have been solved in the new governance tester branch

PR #103, opened 2021-09-23
Governance | initial contract

FloppyDisck:
The initial contract allows users to create proposals and add admin only commands while also having the power to trigger said proposals. The next steps would be to add voting functionality and further improve the rust tests to completely depend on governance


packages/shade_protocol/src/governance.rs
```diff
@@ -0,0 +1,228 @@
...
+use cosmwasm_std::{HumanAddr, Uint128, Binary};
+use secret_toolkit::utils::{InitCallback, HandleCallback, Query};
+use secretcli::secretcli::{TestInit, TestHandle, TestQuery};
+use crate::{
+    asset::Contract,
+    generic_response::ResponseStatus,
+};
+
+// This is used when calling itself
+pub const GOVERNANCE_SELF: &str = "SELF";
```
DrPresident, line 12: This should just be in storage, you can put the address in there at init and then reference it in any context

packages/shade_protocol/src/governance.rs
```diff
@@ -0,0 +1,228 @@
...
+// Admin command variable spot
+pub const ADMIN_COMMAND_VARIABLE: &str = "{}";
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Config {
+    pub admin: HumanAddr,
+    // The amount of time given for each proposal
+    pub proposal_deadline: u64,
+    // The minimum total amount of votes needed to approve deadline
+    pub minimum_votes: Uint128,
```
DrPresident, line 23: Can we rename this to `quorum`

packages/shade_protocol/src/governance.rs
```diff
@@ -0,0 +1,228 @@
...
+    MajorityAgainst,
+    MajorityInFavor,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum Vote {
+    Yes,
+    No,
+    Abstain,
```
DrPresident, line 65: Can these be `For/Against`

packages/shade_protocol/src/governance.rs
```diff
@@ -0,0 +1,228 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum VoteStatus {
+    NoVote,
+    Voting,
+    NotEnoughVotes,
+    MajorityAgainst,
+    MajorityInFavor,
+}
```
DrPresident, line 58: Can this be `ProposalStatus` and have the values like `InProgress, Expired, Accepted, Rejected`, and we would return the actual values of votes so the math on majorities or how many votes are cast/available can be done client side

packages/shade_protocol/src/governance.rs
```diff
@@ -0,0 +1,228 @@
...
+        description: String,
+    },
+    AddAdminCommand {
+        name: String,
+        proposal: String,
+    },
+    RequestRemoveAdminCommand {
+        name: String,
+        description: String,
+    },
```
DrPresident, line 108: Aren't these use-cases covered by the general proposal? e.g. a proposal targeting governance to use `AddAdminCommand` with the desired command

packages/shade_protocol/src/governance.rs
```diff
@@ -0,0 +1,228 @@
...
+        name: String,
+        contract: Contract,
+    },
+
+
+
+    /// Proposal voting
+    Vote {
+        proposal_id: Uint128,
+        option: Vote,
```
DrPresident, line 177: Is this recursive?

packages/shade_protocol/src/governance.rs
FloppyDisck: Theyre more like helper functions, they just build the command for the user

packages/shade_protocol/src/governance.rs
FloppyDisck: We cannot return the actual votes since its supposed to be private

packages/shade_protocol/src/governance.rs
FloppyDisck: that sounds better, ill rename that

packages/shade_protocol/src/governance.rs
FloppyDisck: Agree with the first part, but we should not return votes as the user's vote should be private

packages/shade_protocol/src/governance.rs
FloppyDisck: Remember that its inside an enum so HandleMsg::Vote != Vote

packages/shade_protocol/src/governance.rs
FloppyDisck: Thought about doing this but was unsure if contract has the correct address on init

FloppyDisck: Its still more efficient to just keep it how it is since it avoids serializing & desearializing an already known value

packages/shade_protocol/src/governance.rs
FloppyDisck: Renamed in e416958

packages/shade_protocol/src/governance.rs
FloppyDisck: Renamed in e416958

packages/shade_protocol/src/governance.rs
DrPresident: It definitely has the correct address during msg execution, if that was unreliable we would have alot of problems. It's not being serialized its just being put into storage, its a much better design than having a magic value that convolutes the code.

packages/shade_protocol/src/governance.rs
DrPresident: We aren't returning a users votes, we would be returning the tally for a given proposal (which is necessary for transparency). e.g. querying proposal 34 would give something like `{id: 34, for: 100, against: 50, quorum: 130}`

packages/shade_protocol/src/governance.rs
DrPresident: But we don't have helper functions for building a proposal for any other possible execution, it should be generalized so that we know what is a valid/invalid command (the registered commands), then the user can query those and setup a proposal where we will verify them. Any abstraction or helpers on top of that should be done client side.

packages/shade_protocol/src/governance.rs
FloppyDisck: Oh in that case this has already been discussed, itll be added once voting is added

packages/shade_protocol/src/governance.rs
FloppyDisck: Yeah they were there to make things easier, they can be removed as they dont affect anything if changed

packages/shade_protocol/src/governance.rs
FloppyDisck: Technically it is being serialized, on contract INIT, and then being deserialized every time a proposal aimed at government is written. We could trade off efficiency for readability tho, ill work on it

packages/shade_protocol/src/governance.rs
FloppyDisck: Removed in b405fff and tested in b405fff

DrPresident (approved):

PR #106, opened 2021-09-23
Add PR template

FloppyDisck:
A small template to keep our PRs better organized


mohammedpatla (approved):
I think a basic template works for now. We may need to work on this more in the future if we have more people working into this.

DrPresident (approved):

PR #116, opened 2021-10-10
Airdrop | contract

FloppyDisck:
Added a working airdrop contract, integration tests for it are also implemented


contracts/airdrop/src/handle.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    Ok(HandleResponse {
+        messages: vec![],
+        log: vec![],
+        data: Some( to_binary( &HandleAnswer::UpdateConfig {
+            status: ResponseStatus::Success } )? )
+    })
+}
+
+pub fn try_redeem<S: Storage, A: Api, Q: Querier>(
```
DrPresident, line 45: Can we call this `claim`? I assume its associated pretty directly with `claim_status_r` and I think I like that naming more

contracts/airdrop/src/handle.rs
```diff
@@ -0,0 +1,88 @@
...
+    deps: &mut Extern<S, A, Q>,
+    env: &Env,
+) -> StdResult<HandleResponse> {
+    let config = config_r(&deps.storage).load()?;
+
+    // Check if airdrop started
+    if env.block.time < config.start_date {
```
DrPresident, line 52: For these timestamps can we call them `xxx_time`? at least if they are represented by an epoch. I would love some kind of `DateTime` objects so we could get better day/month/year representations but I don't think thats a priority.

contracts/airdrop/src/handle.rs
```diff
@@ -0,0 +1,88 @@
...
+            return Err(StdError::Unauthorized { backtrace: None })
+        }
+    }
+
+    let key = env.message.sender.to_string();
+
+    // Check if user is eligible
+    if claim_status_r(&deps.storage).load(key.as_bytes())? {
+        return Err(StdError::GenericErr { msg: "Already Claimed".to_string(), backtrace: None })
+    }
```
DrPresident, line 66: Can we do a `may_load` so that a user who was never eligible would get their own message from an eligible user who had already claimed?

contracts/airdrop/src/handle.rs
```diff
@@ -0,0 +1,88 @@
...
+    // Redeem
+    let messages =  vec![mint_msg(env.message.sender.clone(), airdrop.amount,
+                                  None, 1,
+                                  config.airdrop_snip20.code_hash,
+                                  config.airdrop_snip20.address)?];
+
+    // Mark reward as redeemed
+    claim_status_w(&mut deps.storage).update(key.as_bytes(), |claimed| {
+        Ok(true)
+    })?;
```
DrPresident, line 80: Can we also invalidate the `reward` storage? something like set amount to 0

packages/shade_protocol/src/airdrop.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct InitMsg {
+    pub admin: Option<HumanAddr>,
+    pub airdrop_snip20: Contract,
+    // The airdrop time limit
+    pub start_date: Option<u64>,
+    // Can be set to never end
+    pub end_date: Option<u64>,
+    // Secret network delegators snapshot
+    pub rewards: Vec<Reward>,
```
DrPresident, line 27: Can we move `rewards` and `snip20` into a handle like
```
Airdrop {
    snip20,
    Vec<Reward>,
}
```
Then in the storage you could store a users list of assets?
Trying to think of how this contract can still be useful post-airdrop, I believe we are planning to airdrop other tokens in the future such as synthetics

contracts/airdrop/src/handle.rs
FloppyDisck: Id like to keep it as is so theres proof of how much was claimed, maybe we could move it to another storage called claimed amount

packages/shade_protocol/src/airdrop.rs
FloppyDisck: Yeah sounds good, wdym by users list of assets?

contracts/airdrop/src/handle.rs
FloppyDisck: improved in 3c81396

contracts/airdrop/src/handle.rs
FloppyDisck: improved in 3c81396

contracts/airdrop/src/handle.rs
FloppyDisck: improved in 3c81396

FloppyDisck: this should be added once we have the usage flags setup

packages/shade_protocol/src/airdrop.rs
DrPresident: Just if there are multiple overlapping airdrops of different denoms, so one user could query all their pending funds at once. We could also do future airdrops through the staking contract rewards but that would limit

PR #119, opened 2021-10-20
Governance | voting

FloppyDisck:
Added staker based voting into governance


DrPresident (approved):

DrPresident:
I think it would be more clear if we could rename the contract to something like `shd_staking`, so it can be separate from the sSCRT staking and future sATOM etc.

PR #120, opened 2021-10-22
sSCRT staking

DrPresident:
I will be adding unit tests to this as well as further testing for claiming rewards and completed unbondings. I believe those functions will work however it is difficult to generate rewards locally. It's also difficult to test claiming the completed unbonding since it takes 21 days.


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "contracts/snip20"]
 	path = contracts/snip20
 	url = git@github.com:enigmampc/snip20-reference-impl.git
+  branch = master
```
FloppyDisck, line 4: does this fix the module pull issue?

Cargo.toml
```diff
@@ -20,4 +20,5 @@ members = [
 
   # Mock contracts
   "contracts/mock_band",
+  "contracts/scrt_staking",
```
FloppyDisck, line 23: Shouldn't this be in the `Protocol Contracts` section?

packages/shade_protocol/src/scrt.rs
```diff
@@ -0,0 +1,304 @@
+use std::convert::TryFrom;
```
FloppyDisck, line 1: Since the file is more about interacting with the network, could this file be renamed to secret_network or sn_interface?

FloppyDisck (approved):

PR #125, opened 2021-10-24
Shd stake | rewards distribution

FloppyDisck:
Added an efficient way to distribute value across stakers by using the staking model found in Cosmos SDK.
There are now separate unbonding queues per users so they can query their values efficiently. The old unbonding now exists for when we implement treasury rewards distribution.
To allow users to query their stake information, a very simple viewing key system was implemented


chris-ricketts (approved):
Just a couple of nit picks and some Rust tricks you could use if you wanted. 
The unbonding mechanics look sound and efficient.

PR #126, opened 2021-10-24
Automaticaly generate contract schemas in release mode

chris-ricketts:
Adds a build script (`build.rs`) to each contract crate that will generate the CosmWasm schema when the crate is built in release mode. 

If the schemas are already present, it will only re-generate them if any files have changed in `packages/shade_protocol/src`.

I wasn't sure whether to ignore the generated schemas in git or not. Other projects seem to track them in git - possibly to make it easier for other projects to integrate without cloning and generating them.

If we want to track them, I can add them in another commit to this PR.

### DIFF
--- a/contracts/oracle/Cargo.toml
+++ b/contracts/oracle/Cargo.toml
@@ -48,3 +48,7 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
 
 mockall = "0.10.2"
+
+[build-dependencies]
+cosmwasm-schema = { git = "https://github.com/enigmampc/SecretNetwork", tag = "v1.0.4-debug-print" }
+shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol" }
--- a/contracts/airdrop/Cargo.toml
+++ b/contracts/airdrop/Cargo.toml
@@ -46,3 +46,7 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
 mockall = "0.10.2"
 mockall_double = "0.2.0"
+
+[build-dependencies]
+cosmwasm-schema = { git = "https://github.com/enigmampc/SecretNetwork", tag = "v1.0.4-debug-print" }
+shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol" }
--- a/contracts/airdrop/build.rs
+++ b/contracts/airdrop/build.rs
@@ -1,0 +1,30 @@
+use std::path::PathBuf;
+
+use cosmwasm_schema::{export_schema, schema_for};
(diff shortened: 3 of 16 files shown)


contracts/oracle/Cargo.toml
```diff
@@ -48,3 +48,7 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
 
 mockall = "0.10.2"
+
+[build-dependencies]
+cosmwasm-schema = { git = "https://github.com/enigmampc/SecretNetwork", tag = "v1.0.4-debug-print" }
+shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol" }
```
FloppyDisck, line 54: These are already referenced in dependencies, do they also have to be referenced in build-dependencies for the schemas to work?

DrPresident:
> > Having separate message structs would make it easier to document the contract interface. It would also make the code easier to maintain, as adding/removing/modifying a message field would only require additional changes in the specific handler function, rather than changes to pattern matching expressions and function parameters in different files.
> 
> It would make the contract's protocol file look a bit crowded and make calling handle functions from other contracts a bit odd, but if the team prefers the convention we could make an issue to refactor everything later on


We shouldn't change the primary code-base for better fit with tooling/documentation, if we need to do some of it manually because of this that isn't the worst thing in the world. I like the schema generation but could we have it move the nested types (e.g. `Contract`) into the actual attribute its referencing? Assuming the schema is for documentation/client-side use, they don't really need to know/care about the existence of a `Contract` object, just that the field is `contract: {addr, code_hash}`

chris-ricketts:
> We shouldn't change the primary code-base for better fit with tooling/documentation, if we need to do some of it manually because of this that isn't the worst thing in the world. I like the schema generation but could we have it move the nested types (e.g. `Contract`) into the actual attribute its referencing? Assuming the schema is for documentation/client-side use, they don't really need to know/care about the existence of a `Contract` object, just that the field is `contract: {addr, code_hash}`
> 
> I'm not positive on the standard for these schema files, but if we intend for it to represent the actual json sent in the tx could we do `update_config` instead of `UpdateConfigMsg`?

Definitely agree that the protocol interface types (Messages/Responses) should be optimised for simplicity in terms of JSON representation, even if that means documentation is more manual. 

There are quite a few combinations of `serde` attributes we could try to achieve this: https://serde.rs/attributes.html

I wonder if it would be worth creating a separate issue for this rather than bundle it with the schema auto-generation?

chris-ricketts:
Also if this is going to be merged, we need to decide if the schemas are committed to git or ignored

DrPresident:
> Also if this is going to be merged, we need to decide if the schemas are committed to git or ignored

We should commit them assuming they won't be changing outside of actual code changes, I think its good documentation stuff that can live on the repo as long as it is relatively static (I expect they will be)

DrPresident:
Maybe we close this PR until chris is back?

PR #127, opened 2021-10-26
Airdrop | completion flags

FloppyDisck:
Added completion flags for airdrop

The flags can be added before and after deployment, each flag points to an address which will be the contracts that notify the airdrop contract when a user is eligible for that portion of the airdrop

### DIFF
--- a/contracts/airdrop/src/contract.rs
+++ b/contracts/airdrop/src/contract.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{debug_print, to_binary, Api, Binary, Env, Extern, HandleResponse, InitResponse, Querier, StdResult, Storage, Uint128, HumanAddr};
+use cosmwasm_std::{debug_print, to_binary, Api, Binary, Env, Extern, HandleResponse, InitResponse, Querier, StdResult, Storage, Uint128, HumanAddr, StdError};
 use shade_protocol::{
     airdrop::{
         InitMsg, HandleMsg,
@@ -6,21 +6,42 @@ use shade_protocol::{
     }
 };
 use crate::{state::{config_w, reward_w, claim_status_w},
-            handle::{try_update_config, try_claim},
+            handle::{try_update_config, try_add_tasks, try_complete_task, try_claim},
             query };
 use secret_toolkit::snip20::token_info_query;
+use shade_protocol::airdrop::RequiredTask;
 
 pub fn init<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
     env: Env,
     msg: InitMsg,
 ) -> StdResult<InitResponse> {
+
+    // Setup task claim
+    let mut task_claim= vec![RequiredTask {
+        address: env.contract.address,
+        percent: msg.default_claim
+    }];
+    let mut claim = msg.task_claim;
+    task_claim.append(&mut claim);
+
+    // Validate claim percentage
+    let mut count = Uint128::zero();
+    for claim in task_claim.iter() {
+        count += claim.percent;
+    }
+
+    if count > Uint128(100) {
+        return Err(StdError::GenericErr { msg: "tasks above 100%".to_string(), backtrace: None })
+    }
+
     let config = Config{
         admin: match msg.admin {
             None => { env.message.sender.clone() }
             Some(admin) => { admin }
         },
-        airdrop_snip20: msg.airdrop_snip20.clone(),
+        airdrop_snip20: msg.airdrop_token.clone(),
+        task_claim,
         start_date: match msg.start_time {
             None => env.block.time,
             Some(date) => date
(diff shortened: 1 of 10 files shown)


contracts/airdrop/src/query.rs
```diff
@@ -19,8 +19,35 @@ pub fn dates<S: Storage, A: Api, Q: Querier>
...
+    let mut claimed = Uint128::zero();
+    let mut unclaimed = Uint128::zero();
+
+    let config = config_r(&deps.storage).load()?;
+    for (i, task) in config.task_claim.iter().enumerate() {
+        let state = claim_status_r(&deps.storage, i).may_load(key.as_bytes())?;
+
+        match state {
+            None => {}
+            Some(task_claimed) => {
```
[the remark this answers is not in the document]
FloppyDisck: Replaced in a8c4823

contracts/airdrop/src/handle.rs
```diff
@@ -42,6 +42,77 @@ pub fn try_update_config<S: Storage, A: Api, Q: Querier>(
...
+        data: Some( to_binary( &HandleAnswer::AddTask {
+            status: ResponseStatus::Success } )? )
+    })
+}
+
+pub fn try_complete_task<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    env: &Env,
+    address: HumanAddr,
+) -> StdResult<HandleResponse> {
```
DrPresident, line 86: Is this a user that will be marking tasks as "complete"? And do we have an idea of what these tasks might be?

contracts/airdrop/src/contract.rs
```diff
@@ -1,26 +1,47 @@
...
     msg: InitMsg,
 ) -> StdResult<InitResponse> {
+
+    // Setup task claim
+    let mut task_claim= vec![RequiredTask {
+        address: env.contract.address,
+        percent: msg.default_claim
+    }];
+    let mut claim = msg.task_claim;
+    task_claim.append(&mut claim);
```
DrPresident, line 26: So this is the amount you would receive if you were a valid airdrop recipient, but with no tasks completed?

contracts/airdrop/src/contract.rs
```diff
@@ -1,26 +1,47 @@
...
+
+    // Validate claim percentage
+    let mut count = Uint128::zero();
+    for claim in task_claim.iter() {
+        count += claim.percent;
+    }
```
DrPresident, line 32: Probably better to use `Decimal::zero()` for this, otherwise you lose all of your precision. Also this doesn't really need to be a loop since you can only set the default requiredtask and there will only ever be 1 at this point.

contracts/airdrop/src/contract.rs
```diff
@@ -1,26 +1,47 @@
...
+
+    if count > Uint128(100) {
+        return Err(StdError::GenericErr { msg: "tasks above 100%".to_string(), backtrace: None })
+    }
```
DrPresident, line 36: If you switch to decimal this can be 
```
if count > Decimal::one() {
```

contracts/airdrop/src/handle.rs
FloppyDisck: The address of the task is the one that will "complete" the task, this gives us the flexibility to determine what will be the tasks even after the airdrop contract is launched.

Examples could be, conversion minting, staking and governance voting

contracts/airdrop/src/contract.rs
FloppyDisck: Exactly, we can set a default amount.
We want users to be able to use the Shade given to them to unlock the future tasks

contracts/airdrop/src/contract.rs
FloppyDisck: Remember you can initialize the airdrop with pre-set tasks

contracts/airdrop/src/contract.rs
FloppyDisck: You would need to convert everything to decimal for that since Uint128 and Decimal dont match when comparing

DrPresident: But for the purposes of a percentage, integer precision is entirely inadequate and is guaranteed to introduce error into the calculation. It also disallows having any kind of resolution when setting/using these values.

FloppyDisck: Considering the amount is already in uShade, any "decimal" value will still be ignored when claiming so it wouldn't make a difference

DrPresident: I see no reason for this to be integers, the calculation will be inaccurate and there will be lots of error. But if that doesn't matter I will approve.

DrPresident (approved):

contracts/airdrop/src/contract.rs
FloppyDisck: It does matter but decimals doesn't provide a solution for when the number is inevitably converted back into Uint128. Maybe a final calculation once that user's airdrop completion is 100% could give him the possible 1-2 uToken loss

PR #133, opened 2021-11-04
Governance | pre voting

FloppyDisck:
Added proposal funding period to limit proposal spamming. The created proposals are placed in a funding period much like the voting period where users must fund and reach a certain limit, if that limit is reached before the funding deadline the proposal will go into a voting period and return all of the funds to their respective users, if the limit is not reached then it will be sent to treasury (currently not implemented since im waiting for treasury funding handles)

Another thing that was added was more efficient storage for proposal related information, for example votes, amounts, deadlines and status are in their respective storages, this allows for cheaper serialization/deserialization.

### DIFF
--- a/contracts/governance/src/handle.rs
+++ b/contracts/governance/src/handle.rs
@@ -286,7 +286,7 @@ pub fn try_trigger_proposal<S: Storage, A: Api, Q: Querier>(
         run_status = Failure;
     }
     else {
-        run_status = match try_execute_msg(target.unwrap(), proposal.msg.clone()) {
+        run_status = match try_execute_msg(target.unwrap(), proposal.msg) {
             Ok(msg) => {
                 messages.push(msg);
                 Success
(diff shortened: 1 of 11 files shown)


contracts/governance/src/handle.rs
```diff
@@ -98,9 +261,9 @@ pub fn try_trigger_proposal<S: Storage, A: Api, Q: Querier>(
         None => {
             // Check if user is an admin in order to trigger the proposal
             if config.admin == env.message.sender {
-                Ok(Accepted)
+                Ok(Passed)
             } else {
-                Err(StdError::Unauthorized { backtrace: None })
+                Err(StdError::unauthorized())
```
DrPresident, line 266:
```suggestion
                Err(StdError::Unauthorized())
```

DrPresident (approved):

contracts/governance/src/handle.rs
FloppyDisck: StdError::unauthorized() is a function that returns StdError:Unauthorized{ backtrace: None }

Its basically used to shorten the code

PR #134, opened 2021-11-07
Warning cleanup

DrPresident:


FloppyDisck (approved):
Thanks so much for this!

chris-ricketts:
Nice! 

In order to squash the "warning: profiles for the non root package will be ignored, specify profiles at the workspace root:" warnings, you could remove this section from all the nested `Cargo.toml` files and add it underneath the `[workspace]` in the root `Cargo.toml`:
```toml
[profile.release]
opt-level = 3
debug = false
rpath = false
lto = true
debug-assertions = false
codegen-units = 1
panic = 'abort'
incremental = false
overflow-checks = true
```
All the workspace members will then use that profile for release. It would be good to restrict it to the wasm target but that's not possible yet: https://github.com/rust-lang/cargo/issues/4897

PR #138, opened 2021-11-20
Airdrop | minor tweaks

### LINKED ISSUE #128: Add decay
**Is your feature request related to a problem? Please describe.**
Airdrop should have a decay system for the tokens

**Describe the solution you'd like**
We need to decide on either a time based decay system or just dump everything after x amount of time.

For the actual decay airdrop could keep a state of total initial tokens and current awarded tokens, this way calculating total amount to send to treasury should be trivial.

Another think to keep in mind is that if we decide to go by an epoch based decay the total amount to be awarded could be shares, this way reducing total remaining supply of tokens would devalue the individual share value, making it much easier to calculate when redeeming a decayed amount.

The treasury argument (where the decayed value will be dumped) should be optional. The decay dump can only happen if treasury is enabled.

### LINKED ISSUE #129: Replace minting with sending
**Is your feature request related to a problem? Please describe.**
Currently airdrop mints instead of sending

### LINKED ISSUE #135: Airdrop milestone amount precision
**Is your feature request related to a problem? Please describe.**
Depending the airdrop awarded amount users could miss out on potential fractions of the awarded amount.

**Describe the solution you'd like**
Keep a state counter for each user with the total redeemed amount, whenever that user claims from the airdrop if the total award milestones = 100% then return total_awarded - total_redeemed.

**Describe alternatives you've considered**
Using some rounding technique could cause the user to receive more/less that the promised amount

FloppyDisck:
Added a brute decay where the unclaimed amount gets dumped into the ```dump_address``` once ```end_date``` is reached.
Replaced minting items with sending, the contract will now have to be funded before it can start distributing rewards.
Fixed potential issue regarding token distribution when handling certain amounts with certain task percentages; token division will sometimes create very little uToken to not be claimed, to combat this when a user claims and its total percentage achieved is 100% then we just return the current claimed amount subtracted by the total rewarded amount.

Closes #128, Closes #129, Closes #135

### DIFF
--- a/contracts/airdrop/src/handle.rs
+++ b/contracts/airdrop/src/handle.rs
@@ -1,5 +1,5 @@
 use cosmwasm_std::{to_binary, Api, Env, Extern, HandleResponse, Querier, StdError, StdResult, Storage, HumanAddr, Uint128};
-use crate::state::{config_r, config_w, reward_r, claim_status_w, claim_status_r};
+use crate::state::{config_r, config_w, reward_r, claim_status_w, claim_status_r, user_total_claimed_w, total_claimed_w};
 use shade_protocol::airdrop::{HandleAnswer, RequiredTask};
 use shade_protocol::generic_response::ResponseStatus;
 use secret_toolkit::snip20::mint_msg;
@@ -95,7 +95,7 @@ pub fn try_complete_task<S: Storage, A: Api, Q: Querier>(
                         Ok(false)
                     }
                     else {
-                        Err(StdError::Unauthorized { backtrace: None })
+                        Err(StdError::unauthorized())
                     }
                 })?;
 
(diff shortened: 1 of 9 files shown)


contracts/airdrop/src/handle.rs
```diff
@@ -161,4 +184,33 @@ pub fn try_claim<S: Storage, A: Api, Q: Querier>(
...
+
+    // Check if airdrop ended
+    if let Some(end_date) = config.end_date {
+        if let Some(dump_address) = config.dump_address {
+            if env.block.time > end_date {
+                let send_total = (config.airdrop_total - total_claimed_r(&deps.storage).load()?)?;
+                let messages = vec![send_msg(
+                    dump_address, send_total, None, None,
+                    1, config.airdrop_snip20.code_hash,
+                    config.airdrop_snip20.address)?];
```
DrPresident, line 203: if this is just hard-coded like that can we just initialize it in the HandleResponse?

DrPresident (approved):

contracts/airdrop/src/handle.rs
```diff
@@ -161,4 +184,33 @@ pub fn try_claim<S: Storage, A: Api, Q: Querier>(
...
+                    messages,
+                    log: vec![],
+                    data: Some( to_binary( &HandleAnswer::Decay {
+                        status: ResponseStatus::Success } )? )
+                })
+            }
+        }
+    }
+
+    Err(StdError::unauthorized())
```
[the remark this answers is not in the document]
FloppyDisck: all StdErrors have a helper functions to abstract some redundant stuff

contracts/airdrop/src/handle.rs
FloppyDisck: yup we can

FloppyDisck: Fixed in 76ff676

PR #139, opened 2021-11-22
Treasury synthesis

DrPresident:
This is a rough implementation that needs further testing. It is not feature complete and will need further iterations.


contracts/treasury/src/handle.rs
```diff
@@ -1,43 +1,119 @@
...
+
+    allocations_w(&mut deps.storage).update(
+        asset.contract.address.to_string().as_bytes(),
+        |allocs| {
+            let mut alloc_list = match allocs {
+                None => {
+                    vec![]
+                }
+                Some(a) => a,
+            };
```
FloppyDisck, line 55: You can clean up all of the matches like this

```suggestion
            let mut alloc_list = allocs.unwrap_or(vec![]);
```

contracts/treasury/src/handle.rs
```diff
@@ -125,15 +207,209 @@ pub fn try_register_asset<S: Storage, A: Api, Q: Querier>(
...
+    env: &Env,
+    asset: HumanAddr,
+    alloc: Allocation,
 ) -> StdResult<HandleResponse> {
-    let messages = vec![];
+    let config = config_r(&deps.storage).load()?;
+
+    /* ADMIN ONLY */
+    if env.message.sender != config.admin {
+        return Err(StdError::Unauthorized { backtrace: None });
```
FloppyDisck, line 220:
```suggestion
        return Err(StdError::unauthorized());
```

contracts/treasury/src/handle.rs
```diff
@@ -125,15 +207,209 @@ pub fn try_register_asset<S: Storage, A: Api, Q: Querier>(
...
+    }
+
+    let full_asset = match assets_r(&deps.storage).may_load(asset.to_string().as_bytes())? {
+        Some(a) => a,
+        None => {
+            return Err(StdError::GenericErr {
+                msg: "Unregistered asset".to_string(),
+                backtrace: None,
+            });
```
FloppyDisck, line 229: You can make these errors cleaner like this

```suggestion
            return Err(StdError::generic_err ("Unregistered asset"));
```

packages/shade_protocol/src/task.rs
```diff
@@ -0,0 +1,92 @@
+use cosmwasm_std::{HumanAddr, Decimal, Uint128};
```
FloppyDisck, line 1: You can probably generalize this using the admin command style that governance uses. Ill probably work on this in the week to get all of the airdrop stuff out of the way

packages/network_tester/src/scrt_staking.rs
```diff
@@ -0,0 +1,412 @@
...
+        if let micro_mint::QueryAnswer::MintLimit { limit } = query {
+            println!("Mint limit after change request");
+            print_struct(limit);
+        }
+    }
+    */
+
+    Ok(())
+}
+
```
FloppyDisck, line 346: look into ```tests/testnet_integration.rs```. These are now stored in their own functions

packages/network_tester/src/scrt_staking.rs
```diff
@@ -0,0 +1,412 @@
+use colored::*;
```
FloppyDisck, line 1: try moving this into tests and making it a test function so it gets automatically run by the tester

FloppyDisck (approved):
Would be nice if you could look into the comments, but everything looks good aside from that!

packages/shade_protocol/src/task.rs
DrPresident: ugh none of the `task` stuff was supposed to be in there, I was just fiddling around seeing what would work. I'm gonna go cut out all that stuff.

PR #141, opened 2021-12-06
Airdrop account permits

FloppyDisck:


contracts/airdrop/src/handle.rs
```diff
@@ -82,24 +116,159 @@ pub fn try_add_tasks<S: Storage, A: Api, Q: Querier>(
...
+    addresses: Vec<AddressProofPermit>,
+) -> StdResult<HandleResponse> {
+
+    // Check if airdrop active
+    let config = config_r(&deps.storage).load()?;
+
+    // Check that airdrop hasnt ended
+    if !available(&config, env) {
+        return Err(StdError::unauthorized())
+    }
```
DrPresident, line 192: Can this be a `GenericError` with a message indicating the reason?

contracts/airdrop/src/state.rs
```diff
@@ -41,11 +78,51 @@ pub fn total_claimed_w<S: Storage>(storage: &mut S) -> Singleton<S, Uint128> {
...
+}
+
+pub fn account_permit_key_w<S: Storage>(storage: &mut S, account: String) -> Bucket<S, bool> {
+    let key = ACCOUNT_PERMIT_KEY.to_string() + &account;
+    bucket(key.as_bytes(), storage)
+}
+
+pub fn revoke_permit<S: Storage>(storage: &mut S, account: String, permit_key: String) {
+    account_permit_key_w(storage, account).save(permit_key.as_bytes(), &false).unwrap();
+}
```
DrPresident, line 103: For this I think we should be able to revoke all previous permits. In the case where your permit key was compromised / lost it would be important to hard-reset so only the private key holder can re-create themselves a permit.

contracts/airdrop/src/state.rs
```diff
@@ -41,11 +78,51 @@ pub fn total_claimed_w<S: Storage>(storage: &mut S) -> Singleton<S, Uint128> {
...
+
+pub fn is_permit_revoked<S: Storage>(storage: &S, account: String, permit_key: String) -> StdResult<bool> {
+    if account_permit_key_r(storage, account).may_load(permit_key.as_bytes())?.is_some() {
+        Ok(true)
+    }
+    else {
+        Ok(false)
+    }
+}
```
DrPresident, line 112: Instead of this, could we delete revoked permits and tell the user the active permit count?

contracts/airdrop/src/state.rs
FloppyDisck: Permit keys cannot be compromised as they need to be signed. So even if a key is compromised it would not mater as long as the user's mnemonic/account seed is secure.

So even if a key is public, nobody but the account holder can recreate the permit.

I suggest you read into https://github.com/securesecrets/shade/blob/de35976c320a6161718faa869823f50ded571cab/contracts/airdrop/src/state.rs#L127 which is found here: https://github.com/securesecrets/shade/blob/de35976c320a6161718faa869823f50ded571cab/packages/shade_protocol/src/signature/mod.rs#L37

contracts/airdrop/src/state.rs
FloppyDisck: We cannot see the active permit count as this is not stored on-chain since permits are pieces of data that are signed offline.

contracts/airdrop/src/handle.rs
FloppyDisck: Good idea

DrPresident:
I think this would be better to decrypt 1 time when the permit is "initialized" and then just reference that exact permit stored in contract memory to validate a query permit. But this implementation works as-is.

DrPresident (approved):

contracts/airdrop/src/handle.rs
FloppyDisck: Resolved in dc03306

PR #143, opened 2021-12-09
Oracle index

DrPresident:
Initial testing looks good, I can test more thoroughly once we have BAND on pulsar-1. Currently uses a basic weighted average
```
price | weight
1.00  | .5
7.5    | .5
sum((price * weight), ...) / total_weight = price
((1.0 * .5) + (7.5 * .5)) / (.5 + .5) = 4.25
```
```
price | weight
1.00  | .2
7.5    | .5
sum((price * weight), ...) / total_weight = price
((1.0 * .2) + (7.5 * .5)) / (.2 + .5) = 5.642857142857143
```
I used this site to build test cases https://www.rapidtables.com/calc/math/weighted-average-calculator.html
This still needs testing with sswap pairs and such, but using band it is working well

### DIFF
--- a/contracts/oracle/README.md
+++ b/contracts/oracle/README.md
@@ -14,7 +14,7 @@ The oracle contract is used to query the price of different currencies
 ##### Request
 |Name      |Type      |Description                                                                                                        | optional |
 |----------|----------|-------------------------------------------------------------------------------------------------------------------|----------|
-|admin     | string   |  New contract owner; SHOULD be a valid bech32 address, but contracts may use a different naming scheme as well    |  yes     |
+|admin     | string   |  New contract admin; SHOULD be a valid bech32 address, but contracts may use a different naming scheme as well    |  yes     |
 |sscrt     | Contract |  sSCRT snip20 token contract |  no      |
 |band      | Contract |  Band protocol contract   |  no      |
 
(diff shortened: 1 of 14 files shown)


contracts/oracle/README.md
```diff
@@ -14,7 +14,8 @@ The oracle contract is used to query the price of different currencies
 ##### Request
 |Name      |Type      |Description                                                                                                        | optional |
 |----------|----------|-------------------------------------------------------------------------------------------------------------------|----------|
-|owner     | string   |  New contract owner; SHOULD be a valid bech32 address, but contracts may use a different naming scheme as well    |  yes     |
+|admin     | string   |  New contract owner; SHOULD be a valid bech32 address, but contracts may use a different naming scheme as well    |  yes     |
```
FloppyDisck, line 17:
```suggestion
|admin     | string   |  New contract admin; SHOULD be a valid bech32 address, but contracts may use a different naming scheme as well    |  yes     |
```

test-index.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
FloppyDisck, line 1: Since the indexes are self contained equations, could you make unit tests in addition to these integration tests?

FloppyDisck (changes requested):

test-index.py
DrPresident: Basically all the functions require queries to either band or secretswap

FloppyDisck: Could those functions be moved into individual functions so you can spoof that data?

FloppyDisck (approved):

Issue #136, opened 2021-11-16
Governance proposals query by status


FloppyDisck:
**Is your feature request related to a problem? Please describe.**
Implement the remaining query parameter in GetProposals, this only returns the proposals under the given status

PR #144, opened 2021-12-14
[governance] Query proposals by status

LasTshaMAN:
Closes https://github.com/securesecrets/shade/issues/136

Hello guys, recently I've started looking into Rust and thought I'd check out Shade protocol code,
decided to try to implement this feature (seemed like a "good first issue"), so here it is,
if you've got some time for this, any feedback would be welcome :)

Side-note: maybe I should refactor my unit-tests a bit to get rid of duplicated code ?

### DIFF
--- a/contracts/governance/Cargo.toml
+++ b/contracts/governance/Cargo.toml
@@ -32,5 +32,8 @@ shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol" }
 schemars = "0.7"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
+
+[dev-dependencies]
+serde_json = { version = "1.0.67"}
 mockall = "0.10.2"
 mockall_double = "0.2.0"
--- a/contracts/governance/src/query.rs
+++ b/contracts/governance/src/query.rs
@@ -1,5 +1,5 @@
 use cosmwasm_std::{Api, Extern, Querier, StdError, StdResult, Storage, Uint128};
-use shade_protocol::governance::{QueryAnswer, proposal::QueriedProposal};
+use shade_protocol::governance::{QueryAnswer, proposal::QueriedProposal, proposal::ProposalStatus};
 
 use crate::{
     proposal_state::{
@@ -34,7 +34,8 @@ fn build_proposal<S: Storage, A: Api, Q: Querier>(
 pub fn proposals<S: Storage, A: Api, Q: Querier>(
     deps: &Extern<S, A, Q>,
     total: Uint128,
-    start: Uint128) -> StdResult<QueryAnswer> {
+    start: Uint128,
+    status: Option<ProposalStatus>) -> StdResult<QueryAnswer> {
 
     let mut proposals: Vec<QueriedProposal> = vec![];
 
@@ -47,7 +48,15 @@ pub fn proposals<S: Storage, A: Api, Q: Querier>(
     let clamped_start = start.max(Uint128(1));
 
     for i in clamped_start.u128()..((total+clamped_start).min(max).u128() + 1) {
-        proposals.push(build_proposal(&deps, Uint128(i))?)
+        let proposal = build_proposal(&deps, Uint128(i))?;
+
+        // Filter proposal by status if it was specified in fn params.
+        if let Some(s) = &status {
+            if s != &proposal.status {
+                continue
+            }
+        }
+        proposals.push(proposal)
     }
 
     Ok(QueryAnswer::Proposals { proposals })
--- a/contracts/governance/src/query.rs
+++ b/contracts/governance/src/query.rs
@@ -33,8 +33,8 @@ fn build_proposal<S: Storage, A: Api, Q: Querier>(
 
 pub fn proposals<S: Storage, A: Api, Q: Querier>(
     deps: &Extern<S, A, Q>,
-    total: Uint128,
     start: Uint128,
+    end: Uint128,
     status: Option<ProposalStatus>) -> StdResult<QueryAnswer> {
 
     let mut proposals: Vec<QueriedProposal> = vec![];
@@ -47,7 +47,7 @@ pub fn proposals<S: Storage, A: Api, Q: Querier>(
 
     let clamped_start = start.max(Uint128(1));
 
-    for i in clamped_start.u128()..((total+clamped_start).min(max).u128() + 1) {
+    for i in clamped_start.u128()..((end +clamped_start).min(max).u128() + 1) {
         let proposal = build_proposal(&deps, Uint128(i))?;
 
         // Filter proposal by status if it was specified in fn params.
--- a/contracts/governance/src/test.rs
+++ b/contracts/governance/src/test.rs
@@ -35,8 +35,8 @@ mod tests {
 
         // Initialized governance contract has no proposals.
         let res = contract::query(&deps, governance::QueryMsg::GetProposals {
-            total: Uint128(100),
             start: Uint128(0),
+            end: Uint128(100),
             status: Some(ProposalStatus::Funding)
         }).unwrap();
         let value: governance::QueryAnswer = from_binary(&res).unwrap();
@@ -70,8 +70,8 @@ mod tests {
 
         // Should return this proposal when no specific status is specified.
         assert_get_proposals(&deps, governance::QueryMsg::GetProposals {
-            total: Uint128(100),
             start: Uint128(0),
+            end: Uint128(100),
             status: None
         }, |proposals| {
             assert_eq!(1, proposals.len());
@@ -80,8 +80,8 @@ mod tests {
 
         // Should return this proposal when `funding` status is specified.
         assert_get_proposals(&deps, governance::QueryMsg::GetProposals {
-            total: Uint128(100),
             start: Uint128(0),
+            end: Uint128(100),
             status: Some(ProposalStatus::Funding)
         }, |proposals| {
             assert_eq!(1, proposals.len());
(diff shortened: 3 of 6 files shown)


contracts/governance/src/query.rs
```diff
@@ -1,5 +1,7 @@
+use std::os::macos::raw::stat;
 use cosmwasm_std::{Api, Extern, Querier, StdError, StdResult, Storage, Uint128};
-use shade_protocol::governance::{QueryAnswer, QueriedProposal};
+use shade_protocol::governance::{QueryAnswer, QueriedProposal, ProposalStatus};
+use shade_protocol::governance::Vote::No;
```
FloppyDisck, line 4:
```suggestion
use shade_protocol::governance::{QueryAnswer, QueriedProposal, ProposalStatus, Vote::No};
```

contracts/governance/src/query.rs
```diff
@@ -47,7 +50,19 @@ pub fn proposals<S: Storage, A: Api, Q: Querier>(
     let clamped_start = start.max(Uint128(1));
 
     for i in clamped_start.u128()..((total+clamped_start).min(max).u128() + 1) {
```
FloppyDisck, line 52: Because of the addition of status the current way of querying will always return less than the expected amount if status is Some.

I suggest using a while loop and listening for when the proposal's length is either max or total

contracts/governance/Cargo.toml
```diff
@@ -32,5 +32,6 @@ shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol" }
 schemars = "0.7"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
+serde_json = { version = "1.0.67"}
```
FloppyDisck, line 35: Is this import used?

contracts/governance/src/test.rs
```diff
@@ -0,0 +1,114 @@
+#[cfg(test)]
```
FloppyDisck, line 1: We're currently not using these unit tests as they're very cluttered. I would love to teach you about the custom framework we're using.

contracts/governance/src/query.rs
```diff
@@ -47,7 +50,19 @@ pub fn proposals<S: Storage, A: Api, Q: Querier>(
...
+        match &status {
+            Some(s) => {
+                if s != &proposal.status {
+                    continue
+                }
+            },
+            None => (),
+        }
+
+        proposals.push(proposal)
```
DrPresident, line 65: this might be better like 
```
if let Some(s) = &status {
    if s != &proposal.status {
        continue
    }
}
proposals.push(proposal)
```

contracts/governance/src/test.rs
```diff
@@ -0,0 +1,114 @@
...
+    use cosmwasm_std::testing::{mock_dependencies, mock_env};
+    use cosmwasm_std::{HumanAddr, Uint128, coins, from_binary, StdError};
+    use shade_protocol::{snip20::{InitConfig}, snip20, governance, staking,
+                         band, oracle, asset::Contract, airdrop,
+                         airdrop::{Reward, RequiredTask},
+                         governance::{UserVote, Vote, ProposalStatus}, generic_response::ResponseStatus};
+    use crate::contract;
+
+    #[test]
+    fn get_proposals_by_status() {
```
DrPresident, line 13: For this function, could you break these individual tests into their own functions? You could try to parameterize them e.g. https://github.com/securesecrets/shade/blob/main/contracts/oracle/src/test.rs

contracts/governance/Cargo.toml
DrPresident, line 35: Is this used?

DrPresident (commented):
Thanks for the help on this! it looks good just a couple suggestions. Also you should ping @mohammedpatla on telegram.

contracts/governance/Cargo.toml
LasTshaMAN: I'm using it in tests here: https://github.com/securesecrets/shade/pull/144/files#diff-774e1282b450306a1e542cc06f593d6d9f995d6753be1d540c10f2de06b4266fR53

similar to how it is used in `testnet_integration.rs` (used it as example for these unit-tests)

LasTshaMAN: I'm using it in tests here: https://github.com/securesecrets/shade/pull/144/files#diff-774e1282b450306a1e542cc06f593d6d9f995d6753be1d540c10f2de06b4266fR53

similar to how it is used in `testnet_integration.rs` (used it as example for these unit-tests)

contracts/governance/src/query.rs
LasTshaMAN: I'm wondering whether Rust has any code-formatting tools (similar to `goimports` maybe) ?

contracts/governance/src/query.rs
LasTshaMAN: I'm not sure what you mean by "will always return less than the expected amount if status is Some.",

can you elaborate on it ? (I think we can add a test case for this as well),

Does this https://github.com/securesecrets/shade/pull/144#discussion_r769129225 solve that issue ?

contracts/governance/src/test.rs
LasTshaMAN: I've tried to spin up `testnet_integration.rs` but it seems smart contract needs to be compiled into wasm representation first, and I couldn't quite figure it out (based on what docs I found),

do I need to have "special Trusted Hardware" to spin up a working devnet/testnet ?

If you have any docs on this I'd love to learn more :) my Telegram handle should be `iur_ii`, feel free to ping me there (if that's convenient for you)

contracts/governance/src/query.rs
LasTshaMAN: That definitely looks better :) will apply

contracts/governance/src/test.rs
LasTshaMAN: it depends on what's our goal with these tests is,

I've wrote these to "be specific to the feature I'm implementing", not intending to cover whole "GetProposal" functionlaity,
so, convering a couple of check into parametrized test could be an overkill IMO (they seem to be a bit harder to read/write),

but if we want to test "GetProposal" against different inputs (not just "status feature" which this PR is about), then parametrizing probably will be better solution, so I you want me to I can try

LasTshaMAN: @FloppyDisck is mentioning some testing framework here https://github.com/securesecrets/shade/pull/144#discussion_r769128679

maybe we should decide on testing strategy first before proceeding with this ?

contracts/governance/src/query.rs
LasTshaMAN: I guess that's what I'm looking for - https://github.com/intellij-rust/intellij-rust/issues/2158

and maybe `cargo check` / `cargo fix` could help as well ... just wondering what you guys are using ?

contracts/governance/src/test.rs
DrPresident: Agreed, Ideally I think both would be beneficial as these would be "unit tests" vs "integration tests", but I'm also not totally spun up on the other rust tests. I think these tests are good, I would like to see it simplified into a base function with parametrization but I'm also not fond of being too nit-picky about the tests and I think this works fine as-is.

contracts/governance/src/query.rs
FloppyDisck: Lets say theres 100 proposals, my start is 1 and I want a total of 50 values where the ProposalStatus is Accepted. If in the next 50 values theres 5 that are InProgress; then the query will return 45 values instead.

contracts/governance/Cargo.toml
FloppyDisck: You could move it under dev-dependencies since its only used for testing

```suggestion
[dev-dependencies]

serde_json = { version = "1.0.67"}
```

contracts/governance/src/test.rs
FloppyDisck: This is my bad, I failed to provide any good documentation for it.

In the root folder you can ```make``` and that will build all of the contracts. Then just write ```cargo test -- --nocapture --test-threads=1``` and that should be fine.

contracts/governance/src/test.rs
FloppyDisck: I agree with @DrPresident its good to have a balance between both. I usually tend to write them as integration tests when testing the whole contract and use unit tests for individual functions.

contracts/governance/src/query.rs
LasTshaMAN: right, got it,

I guess I was treating `start` and `total` as pagination params, 
meaning the entity requesting proposals can send second request to  grind through proposals 51..100 (possibly getting less than 50 in the result as well),

if it is implemented as you describe, doesn't it kinda break pagination (resulting in duplicates when paginating via `start` and `total`) ?

suppose I want to find **all** proposals in Accepted status, and I have a history of 10 000 proposals (some are in Accepted, others are not),
I can iterate over these in 10 000 / 50 requests each time fetching a bunch (of lenght between 0 and 50) of proposals in Accepted status,
does the implementation you describe support such use-case ? (does it need to ?)

contracts/governance/src/test.rs
LasTshaMAN: I've moved some duplicated code into separate `assert_fn`.

LasTshaMAN:
Squashed all commits into one.

contracts/governance/src/query.rs
FloppyDisck: Good point, I guess it would be best to rename total to something like end to be easier to understand.

FloppyDisck (changes requested):
Everything seems great to me, all that's missing is resolving this comment https://github.com/securesecrets/shade/pull/144/files#r770615444

contracts/governance/src/query.rs
LasTshaMAN: Renamed to `end`, and changed the order of parameters this function (and other similar entities) accepts (so that it reads well).

FloppyDisck (approved):

PR #145, opened 2021-12-15
Airdrop merkle trees

FloppyDisck:
Added merkle proofs to prove airdrop validity. This replaces the high gas cost of uploading the airdrop json

### DIFF
--- a/contracts/airdrop/src/handle.rs
+++ b/contracts/airdrop/src/handle.rs
@@ -1,10 +1,17 @@
-use cosmwasm_std::{to_binary, Api, Env, Extern, HandleResponse, Querier, StdError, StdResult, Storage, HumanAddr, Uint128};
-use crate::state::{config_r, config_w, airdrop_address_r, claim_status_w, claim_status_r, account_total_claimed_w, total_claimed_w, total_claimed_r, account_r, address_in_account_w, account_w, validate_permit, revoke_permit, airdrop_address_w, airdrop_total_w, airdrop_total_r};
-use shade_protocol::{airdrop::{HandleAnswer, Config, claim_info::RequiredTask,
+use cosmwasm_std::{
+    to_binary, Api, Env, Extern, HandleResponse, Querier,
+    StdError, StdResult, Storage, HumanAddr, Uint128, Binary
+};
+use rs_merkle::{Hasher, MerkleProof, algorithms::Sha256};
+use crate::state::{
+    config_r, config_w, claim_status_w, claim_status_r, account_total_claimed_w,
+    total_claimed_w, total_claimed_r, account_r, address_in_account_w, account_w, validate_permit,
+    revoke_permit
+};
+use shade_protocol::{airdrop::{HandleAnswer, Config, claim_info::{RequiredTask, Reward},
                                account::{Account, AddressProofPermit}},
                      generic_response::ResponseStatus};
 use secret_toolkit::snip20::send_msg;
-use shade_protocol::airdrop::claim_info::Reward;
 
 pub fn try_update_config<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
@@ -43,38 +50,6 @@ pub fn try_update_config<S: Storage, A: Api, Q: Querier>(
         messages: vec![],
         log: vec![],
         data: Some( to_binary( &HandleAnswer::UpdateConfig {
-            status: ResponseStatus::Success } )? )
-    })
-}
-
-pub fn try_add_reward_chunk<S: Storage, A: Api, Q: Querier>(
-    deps: &mut Extern<S, A, Q>,
-    env: Env,
-    reward_chunk: Vec<Reward>,
-) -> StdResult<HandleResponse> {
-    let config = config_r(&deps.storage).load()?;
-    // Check if admin
-    if env.message.sender != config.admin {
-        return Err(StdError::unauthorized());
-    }
-
-    // Store the delegators list
-    let mut current_total = Uint128::zero();
-    for reward in reward_chunk {
-        let key = reward.address.to_string();
-
-        airdrop_address_w(&mut deps.storage).save(key.as_bytes(), &reward.amount)?;
-        current_total += reward.amount;
-    }
-
-    airdrop_total_w(&mut deps.storage).update(|total| {
-        Ok(total+current_total)
-    })?;
-
-    Ok(HandleResponse {
-        messages: vec![],
-        log: vec![],
-        data: Some( to_binary( &HandleAnswer::AddRewardChunk {
             status: ResponseStatus::Success } )? )
     })
 }
@@ -120,6 +95,7 @@ pub fn try_create_account<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
     env: &Env,
     addresses: Vec<AddressProofPermit>,
+    partial_tree: Vec<String>,
 ) -> StdResult<HandleResponse> {
 
     let config = config_r(&deps.storage).load()?;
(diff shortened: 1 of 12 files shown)


contracts/airdrop/src/handle.rs
```diff
@@ -210,4 +300,160 @@ pub fn try_decay<S: Storage, A: Api, Q: Querier>(
...
+            }
+        }
+        else {
+            address = sender.clone();
+        }
+
+        // Check that airdrop amount does not exceed maximum
+        if permit.params.amount > config.max_amount {
+            return Err(StdError::generic_err("Amount exceeds maximum amount"))
+        }
```
LasTshaMAN, line 393: this check seems redundant to me, `permit.params.amount` needs to be checked against the exact amount later on, so I'm not sure what it is needed for

contracts/airdrop/src/handle.rs
```diff
@@ -210,4 +300,160 @@ pub fn try_decay<S: Storage, A: Api, Q: Querier>(
...
+    let mut leafs_to_validate: Vec<(usize, [u8; 32])> = vec![];
+
+    // Iterate addresses
+    for permit in addresses.iter() {
+        // Check permit legitimacy / skip if permit is sender
+        let address: HumanAddr;
+        if &permit.params.address != sender {
+            address = validate_permit(storage, permit, config.contract.clone())?;
+            if address != permit.params.address {
+                return Err(StdError::generic_err("Signer address is not the same as the permit address"))
```
LasTshaMAN, line 383: My understanding is this line will never execute because `validate_permit` always returns `permit.params.address` upon successful execution (see screenshot below),

either I'm missing something, or `if address != permit.params.address {` needs to be changed to something else

image

contracts/airdrop/src/handle.rs
```diff
@@ -210,4 +300,160 @@ pub fn try_decay<S: Storage, A: Api, Q: Querier>(
...
+        else {
+            address = sender.clone();
+        }
+
+        // Check that airdrop amount does not exceed maximum
+        if permit.params.amount > config.max_amount {
+            return Err(StdError::generic_err("Amount exceeds maximum amount"))
+        }
+
+        // Check that address has not been added to an account
```
LasTshaMAN, line 395: maybe `// Check that address has not been added to an account` -> `// Check that address has not been added to an account and add it now`

contracts/airdrop/src/handle.rs
```diff
@@ -210,4 +300,160 @@ pub fn try_decay<S: Storage, A: Api, Q: Querier>(
...
+        partial_tree_binary.push(arr);
+    }
+
+    // Prove that user is in airdrop
+    let proof = MerkleProof::<Sha256>::new(partial_tree_binary);
+    // Convert to a fixed length array without messing up the contract
+    let mut root: [u8; 32] = Default::default();
+    root.clone_from_slice(config.merkle_root.as_slice());
+    if !proof.verify(root, &indices, &leaves, config.total_accounts as usize) {
+        return Err(StdError::generic_err("Invalid proof"))
```
LasTshaMAN, line 439: Maybe more details in error message would help with testing/debugging this functionality in the future

contracts/airdrop/src/handle.rs
```diff
@@ -210,4 +300,160 @@ pub fn try_decay<S: Storage, A: Api, Q: Querier>(
...
+
+    total_claimed_w(storage).update(|claimed| {
+        Ok(claimed + redeem_amount)
+    })?;
+
+    Ok((redeem_amount, completed_percentage))
+}
+
+/// Validates all of the information and updates relevant states
+pub fn validate_address_permits<S: Storage>(
```
LasTshaMAN, line 365: It seems this function does a lot of stuff, the part that bothers me is how it modifies `account: &mut Account,`,

I wonder if it can be simplified so that updates to `account` are applied outside of this function ?

packages/shade_protocol/src/airdrop/account.rs
```diff
@@ -0,0 +1,41 @@
...
+pub struct AddressProofMsg {
+    pub address: HumanAddr,
+    // Reward amount
+    pub amount: Uint128,
+    // Used to prevent permits from being used elsewhere
+    pub contract: HumanAddr,
+    // Index of the address in the leafs array
+    pub index: u32,
+    // Used to ban permits
+    pub key: String,
```
LasTshaMAN, line 40: Maybe describing it like this would be more precise:
```rust
    // Can be used by permit owner to revoke permit
    pub revocation_key: String,
```

packages/shade_protocol/src/signature/transaction.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub chain_id: String,
+    /// ignored
+    pub fee: Fee,
+    /// ignored
+    pub memo: String,
+    /// the signed message
+    pub msgs: Vec<TxMsg<T>>,
+    /// ignored
+    pub sequence: Uint128,
+}
```
LasTshaMAN, line 64: it seems there are a bunch of fileds on this struct unrelated to Shade Protocol (e.g. `fee`), I see `ignored` comments there, but why not just remove them ?

contracts/airdrop/src/handle.rs
```diff
@@ -82,24 +91,142 @@ pub fn try_add_tasks<S: Storage, A: Api, Q: Querier>(
...
+    addresses: Vec<AddressProofPermit>,
+    partial_tree: Vec<Binary>,
+) -> StdResult<HandleResponse> {
+
+    let config = config_r(&deps.storage).load()?;
+
+    // Check that airdrop hasnt ended
+    if !available(&config, env) {
+        return Err(StdError::unauthorized())
+    }
```
LasTshaMAN, line 106: maybe a more precise error could be used here

LasTshaMAN (commented):
Just wondering, what the tx price would have been without merkle proofs for uploading the airdrop json ?

contracts/airdrop/src/handle.rs
FloppyDisck: The user must provide a permit for every address, even the current sender, this is for added flexibility (the user can mix and match how many addresses he wants into his account)

The check is used to avoid validating the sender's permit (we already know he owns this address)

contracts/airdrop/src/handle.rs
FloppyDisck: Since we are using merkle trees, a bad actor could find the perfect combination (very improbable) of self owned address + amount to produce the same hash as an already existing leaf. The ```config.max_amount``` is used to limit the potential harm and possible combinations for a leaf node.

contracts/airdrop/src/handle.rs
FloppyDisck: agree, available could be expanded into returning an error instead

packages/shade_protocol/src/signature/transaction.rs
FloppyDisck: Keep in mind this is a signed tx, these fields are important to keep the information up to spec and be compliant with secretd and keplr permit signing.

packages/shade_protocol/src/airdrop/account.rs
FloppyDisck: I disagree, since ```key``` is an identifier for the permit. Maybe the comment above it should clarify this better

contracts/airdrop/src/handle.rs
FloppyDisck: I think it would be best to rename the function to better reflect that it updates accounts, since its a function designed to update the account.

contracts/airdrop/src/handle.rs
```diff
@@ -210,4 +300,160 @@ pub fn try_decay<S: Storage, A: Api, Q: Querier>(
...
+    account: &mut Account,
+    addresses: Vec<AddressProofPermit>,
+    partial_tree: Vec<Binary>,
+) -> StdResult<()> {
+    // Setup the items to validate
+    let mut leafs_to_validate: Vec<(usize, [u8; 32])> = vec![];
+
+    // Iterate addresses
+    for permit in addresses.iter() {
+        // Check permit legitimacy / skip if permit is sender
```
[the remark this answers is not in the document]
FloppyDisck: agree

contracts/airdrop/src/handle.rs
FloppyDisck: ```// update address if its not in an account```

FloppyDisck:
> Just wondering, what the tx price would have been without merkle proofs for uploading the airdrop json ?

Calculations arrived at ~100 gas, which is 10 uSCRT. Its pretty negligible considering that its increase per address is pretty minimal thanks to partial trees.

Something to keep in mind is that the alternative is uploading the whole airdrop JSON which was around 120 scrt

DrPresident (approved):

packages/shade_protocol/src/airdrop/mod.rs
```diff
@@ -1,37 +1,35 @@
...
     pub task_claim: Vec<RequiredTask>,
     // Checks if airdrop has started / ended
     pub start_date: u64,
     pub end_date: Option<u64>,
+    // This is necessary to validate the airdrop information
+    // tree root
+    pub merkle_root: Binary,
+    // tree height
+    pub total_accounts: u32,
+    // max possible reward amount/ used to prevent collision possibility
```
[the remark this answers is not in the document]
DrPresident: For this kind of stuff if you click the "Suggestion" button it gives a quick "commit changes" option

PR #153, opened 2021-12-21
Make format command

LasTshaMAN:
This PR:
- cleans up some artifacts left from SN smart contract generator
- introduces `make format` to apply a common set of formatting rules across all Shade repo (along with `rustfmt.toml` configuration file for it)


makefile
```diff
@@ -20,11 +20,14 @@ all: setup $(CONTRACTS); $(call build_wasm,snip20,snip20_reference_impl)
...
 
 $(compiled_dir) $(checksum_dir):
 	mkdir $@
 
 clean:
 	rm -r $(CONTRACTS)
+
+format:
+	cargo fmt
+	cargo fix --allow-dirty --allow-staged
```
LasTshaMAN, line 33: Not 100% sure about `--allow-dirty --allow-staged` flags, they seem useful to run before committing (this is how I usually use `go fmt`),

could these break my code in unexpected ways ?

DrPresident: Does that basically mean that it will perform formatting on staged changes as well as unstaged changes (dirty)?

chris-ricketts: Just a suggestion but I'd keep the format task just to formatting.

With `cargo fix` its generally good practice to do this within its own commit so the changes can be easily rolled back if they break something.

LasTshaMAN: Have read some more about `cargo fix`, you are right,

since it does more than simple code formatting it is better to run it separately from `make format`, removed it (leaving it up to developer to apply, if he wishes)

DrPresident (approved):

PR #157, opened 2022-01-07
Airdrop bug fixes

FloppyDisck:
Fixed some possible issues with dates when initializing / updating. Added a new query for total claimed.

### DIFF
--- a/contracts/airdrop/Cargo.toml
+++ b/contracts/airdrop/Cargo.toml
@@ -36,3 +36,6 @@ snafu = { version = "0.6.3" }
 rs_merkle = { git = "https://github.com/FloppyDisck/rs-merkle", branch = "node_export" }
 mockall = "0.10.2"
 mockall_double = "0.2.0"
+
+[dev-dependencies]
+flexible-permits = {git = "https://github.com/securesecrets/flexible-permits", tag = "v1.0.0"}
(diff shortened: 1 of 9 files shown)


contracts/airdrop/Cargo.toml
```diff
@@ -36,3 +36,6 @@ snafu = { version = "0.6.3" }
 rs_merkle = { git = "https://github.com/FloppyDisck/rs-merkle", branch = "node_export" }
 mockall = "0.10.2"
 mockall_double = "0.2.0"
+
+[dev-dependencies]
+flexible-permits = {git = "https://github.com/securesecrets/flexible-permits", tag = "v1.0.0"}
```
DrPresident, line 41: is there a reason this isn't under regular dependencies?

DrPresident (approved):

contracts/airdrop/Cargo.toml
FloppyDisck: ```flexible-permits``` gets used inside the ```shade-protocol``` package. The reason it even gets imported in the airdrop contract is because of the tests under ```test.rs```.

DrPresident: OK that makes sense!

PR #158, opened 2022-01-07
Airdrop | query protection

FloppyDisck:
Since total claimed amount is public, malicious nodes can query this information and calculate user claimed amounts. Query now has a rounding or step amount to protect against it. Once the airdrop ```end_date``` is reached and decay amount is claimed then the real amount is exposed.

### DIFF
--- a/contracts/airdrop/src/query.rs
+++ b/contracts/airdrop/src/query.rs
@@ -1,7 +1,7 @@
 use cosmwasm_std::{Api, Extern, Querier, StdResult, Storage, Uint128};
 use shade_protocol::airdrop::{QueryAnswer, account::AccountPermit, claim_info::RequiredTask};
 use crate::handle::decay_factor;
-use crate::state::{config_r, claim_status_r, total_claimed_r, account_r, account_total_claimed_r, validate_account_permit};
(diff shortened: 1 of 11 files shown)


contracts/airdrop/src/query.rs
```diff
@@ -27,8 +27,18 @@ pub fn dates<S: Storage, A: Api, Q: Querier>
...
 (deps: &Extern<S, A, Q>) -> StdResult<QueryAnswer> {
+    let claimed: Uint128;
+    let total_claimed = total_claimed_r(&deps.storage).load()?;
+    if decay_claimed_r(&deps.storage).load()? {
+        claimed = total_claimed;
+    }
+    else {
+        let config = config_r(&deps.storage).load()?;
+        claimed = total_claimed.multiply_ratio(Uint128(1), config.redeem_step_size)
+            .multiply_ratio(config.redeem_step_size, Uint128(1));
```
DrPresident, line 38: Isn't this doing `claimed * (1/r) * (r/1)`
which `(1/r) * (r/1) == 1`?

DrPresident (approved):

contracts/airdrop/src/query.rs
FloppyDisck: Because of Uint rounding it should remove any decimal in the division

DrPresident (approved):

Issue #124, opened 2021-10-23
Add padding to all of the contract handles


FloppyDisck:
**Is your feature request related to a problem? Please describe.**
While handle functions are encrypted and protect what the users are sending, some values can still be predicted or assumed

**Describe the solution you'd like**
To preserve user privacy, all of the handle functions should include an optional padding value to make these TXs harder to predict. A good example is the Snip20 spec.

LasTshaMAN:
Looks like its better to implement this feature once for all contracts (once all of them are in place) ? To lower the chance of missing anything.

FloppyDisck:
Correct, im leaving the issue here for when cleaning up contracts (and closing the remaining issues)

LasTshaMAN:
I'm not that familiar with how Secret Network smart contracts work, but I thought it might be prudent to mention some existing Shade code doesn't use padding when calling other contracts, for example (second parameter is "optional padding" and is `None`):

```rust
    // if burnable then burn if not ignore
    if burning_asset.burnable {
        messages.push(burn_msg(amount, None, 256,
                               burning_asset.contract.code_hash,
                               burning_asset.contract.address)?);
    }
```

Can such sub-calls info be somehow linked to the original method I'm executing on Shade smart contract (with Shade code being open-source) ?

FloppyDisck:
Yup, this also goes under this issue. Basically add padding in our handle functions + snip20 functions that use then

PR #164, opened 2022-01-15
Airdrop | improved privacy

FloppyDisck:
Added handle message and response padding. Fixed a privacy leak regarding claimed total, config was returning the claim amount without any rounding.


LasTshaMAN:
Does this fully resolve https://github.com/securesecrets/shade/issues/124 ? Seems like it does (if not maybe worth clarifying what part of the issue this PR addresses).

DrPresident (approved):

PR #168, opened 2022-01-18
Initializer | improvement

FloppyDisck:
# Description

Because SHD is released before Silk, the initializer would not be very useful. So the SILK init was moved into a separate handle function.

## Notable changes

- Improved how the initializer works
- SILK init is now separate from SHD


contracts/initializer/README.md
```diff
@@ -16,22 +21,77 @@ Contract responsible to initialize the snip20s and keeping the their initial sta
 ##### Request
 |Name             |Type                |Description                   | optional |
 |-----------------|--------------------|------------------------------|----------|
+|admin            | String             | Contract;s admin             |  yes     |
```
rigwild, line 24: Typo

FloppyDisck: oops, let me fix that

PR #171, opened 2022-01-26
changed capture to p*(10^18)

DrPresident:
Slight adjustment to mint capture so it uses 'capture *10^18` for the percentage in the same format as other decimals.

### DIFF
--- a/contracts/micro_mint/src/handle.rs
+++ b/contracts/micro_mint/src/handle.rs
@@ -509,7 +509,7 @@ pub fn calculate_capture(amount: Uint128, capture: Uint128) -> Uint128 {
      * capture_amount = amount * capture / 10000
      */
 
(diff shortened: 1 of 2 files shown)


contracts/micro_mint/src/handle.rs
```diff
@@ -509,7 +509,7 @@ pub fn calculate_capture(amount: Uint128, capture: Uint128) -> Uint128 {
      * capture_amount = amount * capture / 10000
      */
 
-    amount.multiply_ratio(capture, 10000u128)
+    amount.multiply_ratio(capture, 10u128.pow(18))
```
FloppyDisck, line 512: Tests are failing, update `capture_calc` to reflect your latest changes

FloppyDisck (changes requested):

FloppyDisck (approved):

PR #173, opened 2022-01-27
Treasury allowance

DrPresident:
Added monthly allowance to an assets list of allocations. This amount can be refreshed 1x a month, meaning you could refresh on the Jan 30th, then again on Feb 1st. It will query your existing allowance and increase to the monthly amount if it is below, or decrease it to the monthly amount if it is above. Refreshing will refresh all allowances on all assets.

Also added OneTimeAllowance that will increase a given "spender" allowance by an amount given.


contracts/treasury/src/contract.rs
```diff
@@ -17,10 +22,19 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
...
         admin: msg.admin.unwrap_or(env.message.sender.clone()),
+        sscrt: msg.sscrt,
     })?;
 
     viewing_key_w(&mut deps.storage).save(&msg.viewing_key)?;
     self_address_w(&mut deps.storage).save(&env.contract.address)?;
+    asset_list_w(&mut deps.storage).save(&Vec::new())?;
+
+    //init last refresh with epoch 0 so first refresh always goes
+    let timestamp = "0".parse::<i64>().unwrap();
```
FloppyDisck, line 33: why not just write 
```suggestion
    let timestamp = 0;
```

contracts/treasury/src/contract.rs
```diff
@@ -17,10 +22,19 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
 ) -> StdResult<InitResponse> {
     config_w(&mut deps.storage).save(&Config {
         admin: msg.admin.unwrap_or(env.message.sender.clone()),
+        sscrt: msg.sscrt,
     })?;
 
     viewing_key_w(&mut deps.storage).save(&msg.viewing_key)?;
```
FloppyDisck, line 28: you could also read into implementing rng so we never have access to the viewing key 
https://scrt.network/blog/secret-feature-secret-oracles

contracts/treasury/src/state.rs
```diff
@@ -3,15 +3,18 @@ use cosmwasm_storage::{
...
+use shade_protocol::{
+    snip20::Snip20Asset,
+    treasury,
+};
 
 pub static CONFIG_KEY: &[u8] = b"config";
-pub static ASSET_KEY: &[u8] = b"assets";
-pub static ASSET_LIST_KEY: &[u8] = b"asset_list";
+pub static ASSETS: &[u8] = b"assets";
+pub static ASSET_LIST: &[u8] = b"asset_list";
```
FloppyDisck, line 13: can you add the `KEY` suffix to avoid any possible confusion?

makefile
```diff
@@ -3,8 +3,8 @@ compiled_dir=compiled
 checksum_dir=${compiled_dir}/checksum
 
 # Compresses the wasm file, args: compressed_file_name, built_file_name
-define compress_wasm =
-(cd $(contracts_dir)/$(1); cargo unit-test)
+define build =
+(cd $(contracts_dir)/$(1); RUSTFLAGS='-C link-arg=-s' cargo build --release --target wasm32-unknown-unknown --locked --features="debug-print")
```
FloppyDisck, line 7: this build is generally slower than building the whole project once, this is why its called `compress` and called after `build_release` or `build_debug`

FloppyDisck (changes requested):

contracts/treasury/src/state.rs
DrPresident: what confusion? these variables arent used anywhere outside this file

makefile
DrPresident: The makefile as you had changed it wouldn't rebuild the contract on a `make treasury`, I had to rebuild the entire thing to get  my changes into the build.

contracts/treasury/src/contract.rs
DrPresident: All of the balances are public so it doesn't really matter what the viewing key is

FloppyDisck: will treasury be receiving any funds from non shade smart contracts? Cause the viewing key also allows you to check tx history

contracts/treasury/src/state.rs
FloppyDisck: theyre still public, actually why do we even keep our smart contracts keys public, i think we can set them to private in a later PR

makefile
FloppyDisck: Why not use `make` instead? Any contract that has already been built will be skipped. Or generalize the `compress_wasm` and `build_release` to use one contract instead of many

FloppyDisck (changes requested):

makefile
DrPresident: bc make rebuilds everything, it takes way more time and all I need to do is `make treasury`. Without these changes `make treasury` does basically nothing except re-run unit-tests on whatever build is already there, which doesn't make sense

contracts/treasury/src/contract.rs
DrPresident: It definitely has the potential to receive from many different smart contracts, as well as any configured snip-20. If we need the viewing key to be more secure we can set it to something more complex at init. It could generate something itself but thats not really relevant to this PR and something that could be added later.

makefile
FloppyDisck: Could you atleast fix it so it doesnt take longer when runningmake? For example, removing --features="debug-print" and preventing it from compiling individual contracts after build_release or build_debug is called.

FloppyDisck (approved):

PR #174, opened 2022-02-02
Tester improvement

FloppyDisck:
# Description

Added a way for testers to also create a report, currently this report only includes the gas costs but it can be expanded into showing more relevant data regarding the MSg

## Notable changes

- Removed unused `secretcli` functions
- Improved existing `secretcli` functions
- Added gas report feature to `secretcli`
- Improved merkle tree height calculation using a custom log2 ceil method
- Removed redundant code in the airdrop


packages/network_integration/src/contract_helpers/minter.rs
```diff
@@ -1,24 +1,26 @@
 use crate::{
-    contract_helpers::governance::{create_and_trigger_proposal, get_contract, init_contract},
+    contract_helpers::governance::{create_and_trigger_proposal, get_contract, init_with_gov},
     utils::{
         print_contract, print_epoch_info, print_header, print_vec, GAS, MICRO_MINT_FILE, VIEW_KEY,
```
DrPresident, line 4: I made the `micro_mint -> mint` changes in https://github.com/securesecrets/shade/pull/175, can we get that merged first? I adjusted these test files so it shouldn't be too disruptive here

FloppyDisck: sounds good

DrPresident (approved):

packages/network_integration/src/utils.rs
```diff
@@ -48,7 +48,7 @@ pub fn print_epoch_info(minter: &NetContract) {
     println!("\tEpoch information");
     let msg = mint::QueryMsg::GetMintLimit {};
 
-    let query: micro_mint::QueryAnswer = query(minter, &msg, None).unwrap();
+    let query: mint::QueryAnswer = query_contract(minter, &msg).unwrap();
```
FloppyDisck, line 51: @DrPresident Could you change to `let query: mint::QueryAnswer = query(minter, &msg, None).unwrap();`

PR #178, opened 2022-02-11
Smart Contract Errors

FloppyDisck:
# Description
The front end team thought it would be good to have easier to parse errors, this solves that issue by giving each error a code ID, name, description and variables used in description

## Notable changes

- Improved error verbosity

## Next steps

- Implement in all of the other contracts
- Implement on any code that might contain an error; unexpected errors are fine


contracts/airdrop/src/handle.rs
```diff
@@ -50,27 +51,43 @@ pub fn try_update_config<S: Storage, A: Api, Q: Querier>(
...
             if let Some(end_date) = end_date {
                 if start_date > end_date {
-                    return Err(StdError::generic_err(
-                        "New start date is greater than end date",
+                    return Err(invalid_dates(
+                        "EndDate",
+                        end_date.to_string().as_str(),
+                        "before",
+                        "StartDate",
+                        start_date.to_string().as_str()
```
DrPresident, line 59: Can this be just 1 string using the format?

packages/network_integration/src/utils.rs
```diff
@@ -46,15 +46,15 @@ pub fn print_contract(contract: &NetContract) {
...
-    if let mint::QueryAnswer::MintLimit { limit } = query {
-        println!(
-            "\tFrequency: {}\n\tCapacity: {}\n\tTotal Minted: {}\n\tNext Epoch: {}",
-            limit.frequency, limit.mint_capacity, limit.total_minted, limit.next_epoch
-        );
+    if let mint::QueryAnswer::Limit { limit, minted, last_refresh} = query {
+        // println!(
+        //     "\tFrequency: {}\n\tCapacity: {}\n\tTotal Minted: {}\n\tNext Epoch: {}",
+        //     limit.frequency, limit.mint_capacity, limit.total_minted, limit.next_epoch
+        // );
```
DrPresident, line 57: its just tests so leave it in if you want but it seems this block does nothing

DrPresident (approved):

contracts/airdrop/src/handle.rs
FloppyDisck: i wanted to make it flexible, so i also use it for decay date and all that. Thats why I also specify the variable's name

packages/network_integration/src/utils.rs
FloppyDisck: I commented it out since I used to use that function for some tests but since mint has changed when itll be fixed in a later PR

PR #182, opened 2022-02-16
Added viewing keys

FloppyDisck:
# Description

Implements secure viewing keys from `query-authentication` just in case permits result in issues with the high user query load

## Notable changes

- Added viewing keys

### DIFF
--- a/contracts/airdrop/src/query.rs
+++ b/contracts/airdrop/src/query.rs
@@ -1,4 +1,4 @@
-use crate::state::{address_in_account_r, validate_address_permit};
+use crate::state::{account_viewkey_r, address_in_account_r, validate_address_permit};
 use crate::{
     handle::decay_factor,
     state::{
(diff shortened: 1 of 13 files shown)


contracts/airdrop/src/query.rs
```diff
@@ -106,3 +104,39 @@ pub fn account<S: Storage, A: Api, Q: Querier>(
...
+    account: HumanAddr,
+    key: String,
+    current_date: Option<u64>,
+) -> StdResult<QueryAnswer> {
+
+    // Validate address
+    let stored_hash = account_viewkey_r(&deps.storage).load(
+        account.to_string().as_bytes())?;
+
+    if !AccountKey(key).compare(&stored_hash) {
```
DrPresident, line 133:
```suggestion
    if !AccountKey(key).hash().compare(&stored_hash) {
```

DrPresident (approved):

contracts/airdrop/src/query.rs
FloppyDisck: `.compare()` hashes `AccountKey` for you

PR #196, opened 2022-03-21
Staking | Interface

FloppyDisck:
# Description

Improves shade staking and adds more flexibility to the governance contract

## Notable changes

- Added SHD staking
- Implemented a way to have liquid staking
- Balances can be reflected for voting
- Transfer cooldown to tackle cheap and efficient voting
- Added some lockdown features in case of exploit

## Next steps

- Create docs page
- Add errors library support
- Update current `Uint256` math to cosmwasm's library
- Change sub-module branch to `main`


.gitmodules
```diff
@@ -1,3 +1,8 @@
 [submodule "contracts/snip20"]
 	path = contracts/snip20
 	url = https://github.com/scrtlabs/snip20-reference-impl.git
+    branch = master
+[submodule "contracts/shd_staking"]
+	path = contracts/shd_staking
+	url = https://github.com/securesecrets/SPIP-STKN-0
```
DrPresident, line 7: It looks like this is just a clone of the snip20 implementation? Why does it need to be a submodule and not just live in the repo?

DrPresident (approved):

PR #198, opened 2022-04-05
fixed overflow error

DrPresident:
Moved `eval_index` onto the compat Uint512 to avoid overflow errors

### DIFF
--- a/contracts/oracle/Cargo.toml
+++ b/contracts/oracle/Cargo.toml
@@ -37,3 +37,4 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
 
 mockall = "0.10.2"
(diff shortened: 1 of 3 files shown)


DrPresident:
> Why not use `Uint256` instead of `Uint512`?

Uint512 is bigger and I don't know of any benefits to using 256 instead. I'd rather account for large baskets for indexes and thats going to require handling bigger numbers.

FloppyDisck:
> > Why not use `Uint256` instead of `Uint512`?
> 
> Uint512 is bigger and I don't know of any benefits to using 256 instead. I'd rather account for large baskets for indexes and thats going to require handling bigger numbers.

256 is smaller so it should spend less gas, but that can be studied later when were improving gas consumption

FloppyDisck (approved):

PR #199, opened 2022-04-06
Mint fix

DrPresident:
Fixed some things in mint, I don't think anything was really broken but its slightly cheaper now and removed some redundant code

### DIFF
--- a/contracts/mint/src/handle.rs
+++ b/contracts/mint/src/handle.rs
@@ -206,8 +206,10 @@ pub fn try_limit_refresh<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
     env: Env,
     limit: Limit,
-) -> StdResult<()> {
+) -> StdResult<Uint128> {
+
     match DateTime::parse_from_rfc3339(&limit_refresh_r(&deps.storage).load()?) {
+
         Ok(parsed) => {
             let naive = NaiveDateTime::from_timestamp(env.block.time as i64, 0);
             let now: DateTime<Utc> = DateTime::from_utc(naive, Utc);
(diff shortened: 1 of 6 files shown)


contracts/mint/src/handle.rs
```diff
@@ -124,17 +125,30 @@ pub fn try_burn<S: Storage, A: Api, Q: Querier>(
...
-        if token_config.burn_enabled {
-            messages.push(burn_msg(
-                burn_amount,
-                None,
-                None,
-                256,
-                burn_asset.asset.contract.code_hash.clone(),
-                burn_asset.asset.contract.address.clone(),
-            )?);
+    if burn_amount > Uint128::zero() {
```
FloppyDisck, line 128: Duplicate code, can be reduced to

```if burn_amount > Uint128::zero() {
        // Try to burn
        if let Some(token_config) = &burn_asset.asset.token_config {
            if token_config.burn_enabled {
                messages.push(burn_msg(
                    burn_amount,
                    None,
                    None,
                    256,
                    burn_asset.asset.contract.code_hash.clone(),
                    burn_asset.asset.contract.address.clone(),
                )?);
            }
        } else if let Some(recipient) = config.secondary_burn {
            messages.push(send_msg(
                recipient,
                burn_amount,
                None,
                None,
                None,
                1,
                burn_asset.asset.contract.code_hash.clone(),
                burn_asset.asset.contract.address.clone(),
            )?);
        }
    }```

**Review by FloppyDisck — CHANGES_REQUESTED** [review]

File: contracts/mint/src/handle.rs (hunk 1 above)
Comment (DrPresident) [reply contracts/mint/src/handle.rs:128]: It isn't duplicated, there are 3 cases to be accounted for.
```
if has_token_config() {
   if burn_enabled => do_burn()
   else => do_secondary_burn()
}
else if no_token_confg() {
    do_secondary_burn()
}
```

FloppyDisck (approved):

PR #237, opened 2022-07-25
Bonds audit

wahlbergkyle:
# Description

Responding to Certik audit of Bonds contract, as well as adding fix to allow for stkd-SCRT compatibility.

## Notable changes

- User account created based off of "from", not "sender" when assets are deposited
- fetch_snip20 now won't error out if token_config DNE
- HandleResponse from update_config and update_limit_config improved


wahlbergkyle:
Test failed in the same way "main" currently is. This only concerns bonds audit commits, hence that failure remains.

SissonJ (approved):

FloppyDisck:
This should be merged into dev and then into main

FloppyDisck (approved):

PR #241, opened 2022-08-02
Sky migration to cosmwasm v1

SissonJ:
Migrating sky to cw1

### DIFF
--- a/contracts/sky/Cargo.toml
+++ b/contracts/sky/Cargo.toml
@@ -24,7 +24,7 @@ default = []
 backtraces = ["shade-protocol/backtraces"]
 
 [dependencies]
-shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol", features = ["sky", "sky-impl", "adapter", "math", "shade_admin"] }
+shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol", features = ["sky", "sky-utils", "adapter", "math", "shade_admin"] }
 #shade-admin = { git = "https://github.com/securesecrets/shadeadmin", branch = "cosmwasm-v1-refactor", optional = true }
 schemars = "0.7"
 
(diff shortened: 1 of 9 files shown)


contracts/sky/Cargo.toml
```diff
@@ -24,7 +24,8 @@ default = []
 backtraces = ["shade-protocol/backtraces"]
 
 [dependencies]
-shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol", features = ["sky", "sky-impl", "adapter", "math"] }
+shade-protocol = { version = "0.1.0", path = "../../packages/shade_protocol", features = ["sky", "sky-impl", "adapter", "math", "shade_admin"] }
```
FloppyDisck, line 27: you can make `sky-impl` import all the necessary features such as `sky` and `shade_admin` so you only have to import the impl feature and done

packages/shade_protocol/Cargo.toml
```diff
@@ -88,7 +88,7 @@ backtraces = ["cosmwasm-std/backtraces"]
 debug-print = [] # TODO: remove this from all cargo configs
 
 [dependencies]
-cosmwasm-std = { git = "https://github.com/scrtlabs/cosmwasm", branch = "secret" }
+cosmwasm-std = { git = "https://github.com/scrtlabs/cosmwasm", branch = "secret", features = ["staking"] }
```
FloppyDisck, line 91: you should leave the staking feature out and import it whereever you need it

contracts/sky/Cargo.toml
SissonJ: Sky, sky-impl, and shade-admin should remain separate since eventually will refactor sky-impl into a dex helper

FloppyDisck: yes but sky-impl relies on the sky feature

sky acts like the interface and the impl is functionality, if you cant import the impl without importing sky then they should be automatically imported

FloppyDisck (approved):

PR #242, opened 2022-08-03
Added a way to reduce the annoyance of using newtypes with storage traits

FloppyDisck:
# Description

This addition aims to clean up how newtype storage is imagined

Originally, you would have this issue:
```
#[cw_serde]
pub struct NewType(pub u64);

impl ItemStorage for NewType {
    const ITEM: Item<'static, Self> = Item::new("newtype-");
}
```
Which would translate to using it like so
```
let mut t = NewType::load(deps.storage())?.0;
t = 200
NewType(t)::save(deps.storage())?;
```

This addition makes it so you can  change how the newtypes are seend, instead of being a storage wrapper for the item, it works like a boilerplate code generator
```
#[cw_serde]
pub struct NewType;

impl GenericItemStorage for NewType {
    const ITEM: Item<'static, u64> = Item::new("newtype-");
}
```
Which would translate to using it like so
```
let mut t = NewType::load(deps.storage())?;
t = 200
NewType::save(deps.storage(), t)?;
```

### DIFF
--- a/packages/shade_protocol/src/contract_interfaces/snip20/helpers.rs
+++ b/packages/shade_protocol/src/contract_interfaces/snip20/helpers.rs
@@ -82,7 +82,7 @@ pub fn mint_msg(
     contract: &Contract
 ) -> StdResult<CosmosMsg> {
     ExecuteMsg::Mint {
-        recipient: recipient,
+        recipient,
         amount,
         memo,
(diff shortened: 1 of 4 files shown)


hydropump3 (approved):
comments in slack were already answered so it looks good

PR #248, opened 2022-08-10
Peg stability

SissonJ:
# Description

## Notable changes

- Added a peg stability contract
- NO TESTS

## Next steps

-Tests


contracts/peg_stability/src/contract.rs
```diff
@@ -0,0 +1,89 @@
...
+};
+
+pub fn instantiate(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    msg: InstantiateMsg,
+) -> StdResult<Response> {
+    let config = Config {
+        shd_admin: msg.shd_admin.clone(),
```
DrPresident, line 31: I think the current usages mostly call this `admin_auth`

contracts/peg_stability/src/contract.rs
```diff
@@ -0,0 +1,89 @@
...
+        treasury: msg.treasury.clone(),
+        symbols: vec![],
+        payback: msg.payback,
+        self_addr: env.contract.address.clone(),
+        dump_contract: msg.dump_contract,
+    };
+    config.save(deps.storage)?;
+    ViewingKey::save(deps.storage, &msg.viewing_key.clone())?;
+    Ok(Response::new()
+        .add_submessage(SubMsg::new(set_viewing_key_msg(
```
DrPresident, line 44: Probably doesn't matter but if you aren't using the response probably better to just use `add_message`

contracts/peg_stability/src/handle.rs
```diff
@@ -0,0 +1,238 @@
...
+) -> StdResult<Response> {
+    //Admin-only
+    let mut config = Config::load(deps.storage)?;
+    validate_admin(
+        &deps.querier,
+        env.contract.address.to_string(),
+        info.sender.to_string(),
+        &config.shd_admin,
+    )?;
+    if pairs.len() < 1 {
```
DrPresident, line 93:
```suggestion
    if pairs.is_empty() {
```

contracts/peg_stability/src/handle.rs
```diff
@@ -0,0 +1,238 @@
...
+        return Err(StdError::generic_err("Must pass at least one pair"));
+    }
+    let res0: TokenInfo = fetch_snip20(&pairs[0].token0, &deps.querier)?.token_info;
+    let res1: TokenInfo = fetch_snip20(&pairs[0].token1, &deps.querier)?.token_info;
```
DrPresident, line 97: This will query the token config, I would directly use `snip20::helpers::token_info`, I would also rename the variables to be more meaningful like `token0_info`

contracts/peg_stability/src/handle.rs
```diff
@@ -0,0 +1,238 @@
...
+    let other_asset;
+    if config.snip20 == pairs[0].token0 {
+        config.symbols = vec![res0.symbol, res1.symbol];
+        other_asset = pairs[0].token1.clone();
+    } else {
+        config.symbols = vec![res1.symbol, res0.symbol];
+        other_asset = pairs[0].token0.clone();
+    }
```
DrPresident, line 105: Might be good to add a final `else` to fail out in the worst (likely unreachable) case

contracts/peg_stability/src/handle.rs
```diff
@@ -0,0 +1,238 @@
...
+    env: Env,
+    info: MessageInfo,
+    pairs: Vec<ArbPair>,
+) -> StdResult<Response> {
+    let mut config = Config::load(deps.storage)?;
+    if config.pairs.len() == 0 {
+        return Ok(try_set_pairs(deps, env, info, pairs)?);
+    } else if pairs.len() < 1 {
+        return Err(StdError::generic_err("Must pass at least 1 pair"));
+    }
```
DrPresident, line 138: I would do `if config.pairs.is_empty()` to fail, currently the two checks are identical -- `len == 0` or `len < 1`

contracts/peg_stability/src/handle.rs
```diff
@@ -0,0 +1,238 @@
...
+            status: true,
+        })?),
+    )
+}
+
+pub fn try_remove_pair(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    index: Uint128,
```
DrPresident, line 177: Can we do the remove based on pair addr?

FloppyDisck (approved):

contracts/peg_stability/src/handle.rs
```diff
@@ -0,0 +1,238 @@
...
+        &deps.querier,
+        env.contract.address.to_string(),
+        info.sender.to_string(),
+        &config.shd_admin,
+    )?;
+    let i = index.u128() as usize;
+    if config.pairs.len() == 0 {
+        return Err(StdError::generic_err("No pairs to remove"));
+    }
+    if i > config.pairs.len() - 1 {
```
DrPresident, line 191:
```suggestion
    if i >= config.pairs.len() {
```

contracts/peg_stability/src/query.rs
```diff
@@ -0,0 +1,204 @@
...
+        config: Config::load(deps.storage)?,
+    })
+}
+
+pub fn get_balance(deps: Deps) -> StdResult<QueryAnswer> {
+    let viewing_key = ViewingKey::load(deps.storage)?;
+    let config = Config::load(deps.storage)?;
+
+    let mut res = snip20::QueryMsg::Balance {
+        address: config.self_addr.clone().to_string(),
```
DrPresident, line 29: I would move `self_addr` into its own storage as nobody would ever need to see it in the config when querying, and it isn't configurable

contracts/peg_stability/src/query.rs
```diff
@@ -0,0 +1,204 @@
...
+        key: viewing_key,
+    }
+    .query(&deps.querier, &config.snip20)?;
+
+    match res {
+        snip20::QueryAnswer::Balance { amount } => Ok(QueryAnswer::Balance { snip20_bal: amount }),
+        _ => Ok(QueryAnswer::Balance {
+            snip20_bal: Uint128::zero(),
+        }),
+    }
```
DrPresident, line 39: This should handle the error case, as well as return `amount` instead of `0`

contracts/peg_stability/src/query.rs
```diff
@@ -0,0 +1,204 @@
...
+    /*let res: Vec<OraclePrice> = router::QueryMsg::GetPrices {
+        keys: config.symbols,
+    }
+    .query(&deps.querier, &config.oracle)?
+    .prices;
+    let prices = vec![
+        Uint128::new(res[0].data.rate.u128()),
+        Uint128::new(res[1].data.rate.u128()),
+    ];*/
+    let prices = vec![Uint128::zero(), Uint128::zero()];
```
DrPresident, line 70: I believe there is syntax for `vec![Uint128::zero(); 2]` to recreate this behavior

contracts/peg_stability/src/query.rs
```diff
@@ -0,0 +1,204 @@
...
+                prices[0],
+                prices[1],
+            );
+        } else {
+            temp = calculate_swap_amount(
+                t1_amount.checked_mul(Uint128::new(10).pow(18 - snip20_dec.clone()))?,
+                t0_amount.checked_mul(Uint128::new(10).pow(18 - other_dec.clone()))?,
+                prices[0],
+                prices[1],
+            );
```
DrPresident, line 98: should these `prices[i]` be swapped?

contracts/peg_stability/src/query.rs
```diff
@@ -0,0 +1,204 @@
...
+fn calculate_swap_amount(
+    poolsell: Uint128,
+    poolbuy: Uint128,
+    pricesell: Uint128,
+    pricebuy: Uint128,
+) -> Uint128 {
+    let nom = Uint256::from(pricebuy.isqrt())
+        * Uint256::from(poolbuy.isqrt())
+        * Uint256::from(poolsell.isqrt());
+    let denom = Uint256::from(pricesell.isqrt());
```
DrPresident, line 158: The variable name `denom` kind of implies a kind of token, as opposed to a token amount/price. Also you are doing the `Uint256::from(pricesell.isqrt())` 2x

DrPresident (approved):

PR #254, opened 2022-08-15
Admin migration

GumboLimbo:
Merged admin auth into shade repo, rolled back to preferred version of admin auth, wrote additional unit tests


packages/multi_test/src/multi.rs
```diff
@@ -87,7 +88,6 @@ pub mod query_auth {
 //     multi_derive::implement_multi!(Treasury, treasury);
 // }
 
-#[cfg(feature = "peg_stability")]
```
SissonJ, line 90: Why are we removing this feature flag?

packages/shade_protocol/Cargo.toml
```diff
@@ -27,6 +27,7 @@ query_auth_lib = ["dep:query-authentication"]
 multi-test = ["dep:secret-multi-test", "dep:anyhow", "interface"]
 ensemble = ["dep:fadroma"]
 
+# Implementing new contracts:
```
SissonJ, line 30: Duplicated line

SissonJ, line 30:
```suggestion
```

SissonJ (approved):

PR #259, opened 2022-08-17
Dao test jbs

SissonJ:
# Description

## Notable changes

- Added new contract query

## Next steps


contracts/treasury/src/execute.rs
```diff
@@ -297,16 +310,24 @@ pub fn rebalance(deps: DepsMut, env: &Env, info: MessageInfo, asset: Addr) -> St
...
-    println!("portion total {}", portion_total);
-    println!("amount_total {}", amount_total);
+    println!(
+        "SUS AFTER GREATER {} - {}",
+        token_balance.u128(),
+        amount_total.u128(),
+    );
 
     for allowance in portions {
+        portion_total += token_balance - amount_total;
```
DrPresident, line 328: I don't think this makes sense, this number is going to get bigger every iteration of the loop and I think its only working because you are likely doing only 1 portion allocation

SissonJ: It fails out when I have 0 portion allocations

DrPresident (approved):

PR #264, opened 2022-08-24
more more more tests + some tm unbond upgrades

SissonJ:
# Description

## Notable changes

- Added new contract query

## Next steps


packages/multi_test/src/interfaces/dao.rs
```diff
@@ -201,6 +202,56 @@ pub fn system_balance(
...
 }
 
+/*pub fn add_or_sub_adapter_tokens(
+    chain: &mut App,
+    sender: &str,
+    is_add: bool,
+    adapter_ids: Vec<usize>,
+    amount_to_add: Uint128,
+    num_managers: usize,
+) -> StdResult<()> {
```
DrPresident, line 212: Can we delete this or do you think it may be useful moving forward?

DrPresident (approved):
Can we get the metrics added back into this function?

PR #298, opened 2022-09-08
Make overhaul

FloppyDisck:
# Description

Replaces the hard to read makefile with an rustified version.

`cargo install --force cargo-make`
To build the whole project for release
`cargo make  build-all`
Release just one package
`cargo make release governance`
To test a package
`cargo make test governance`
Or all packages
`cargo make test-all`

## Notable changes

- Added rust cargo make support

## Next steps

- Improve certain setups, like docker support and automating launching a contract withing the makefile


DrPresident (commented):
My biggest concern with this is changing the command from `make` to `cargo make --makefile ../Makefile.toml build-all`, it doesn't seem to add any necessary functionality but makes  the command way longer.

FloppyDisck:
> My biggest concern with this is changing the command from `make` to `cargo make --makefile ../Makefile.toml build-all`, it doesn't seem to add any necessary functionality but makes the command way longer.

Cargo make is a much easier setup to expand, on top of this it has workflows support so its way easier to have functional and less messy testers. Also, apparently a new version allows you to run `cargo make build-all` and it works just fine, I just tested it.

DrPresident (approved):
I think the additions in `overseer` may have been accidental and should probably removed, but looks good otherwise. 
I think we should eventually use the `makefile` to call into this

FloppyDisck:
Fully agree with the makefile wrapping the cargo make

PR #309, opened 2022-09-13
Treasury Manager Query and Warnings Cleanup

SissonJ:
Added query tests for treasury manager and cleaned up a lot of warnings


FloppyDisck:
could the PR be more verbose? This is seen publicly on the git history

PR #310, opened 2022-09-14
governance/number-size-fix

FloppyDisck:
# Description

Replaced many of the u128 references with more logical types such as u16 and u32 to conserve gas.

## Notable changes

- Lowered gas prices

### DIFF
--- a/contracts/governance/src/handle/assembly_msg.rs
+++ b/contracts/governance/src/handle/assembly_msg.rs
@@ -32,7 +32,7 @@ pub fn try_add_assembly_msg(
     info: MessageInfo,
     name: String,
     msg: String,
-    assemblies: Vec<Uint128>,
+    assemblies: Vec<u16>,
 ) -> StdResult<Response> {
     let id = ID::add_assembly_msg(deps.storage)?;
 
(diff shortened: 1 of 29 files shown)


contracts/governance/src/handle/assembly_msg.rs
```diff
@@ -32,7 +32,7 @@ pub fn try_add_assembly_msg(
     info: MessageInfo,
     name: String,
     msg: String,
-    assemblies: Vec<Uint128>,
+    assemblies: Vec<u16>,
```
GumboLimbo, line 35: Is the u16 here implicitly limiting the max number of assemblies to 65500ish? That is a high number but I want to make sure it's been considered. If this is the case we should document this in public documentation somehow so nobody gets surprised by it down the road

GumboLimbo (approved):

contracts/governance/src/handle/assembly_msg.rs
FloppyDisck: All governance types were lowered in size since we're never going to realistically go above ~50 assemblies. I still think the type sizes can be lowered even more but I dont wanna risk it with a `u8` type.

PR #313, opened 2022-09-19
generated label should now have more entropy do accurately attack

FloppyDisck:
# Description

Generated labels when migrating should be harder to predict

### DIFF
--- a/contracts/governance/src/handle/migration.rs
+++ b/contracts/governance/src/handle/migration.rs
@@ -29,7 +29,7 @@ pub fn try_migrate(
     env: Env,
     _info: MessageInfo,
     id: u64,
-    label: String,
+    mut label: String,
     code_hash: String,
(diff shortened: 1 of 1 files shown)


GumboLimbo (approved):

contracts/governance/src/handle/migration.rs
```diff
@@ -39,6 +39,8 @@ pub fn try_migrate(
 
     RuntimeState::Migrated {}.save(deps.storage)?;
 
+    label.push_str(&env.block.time.nanos().to_string());
```
GumboLimbo, line 42: Is it possible to use a nonce here?

FloppyDisck: It is, but you'd need entropy which governance doesnt have, the only possible entropy to use would be the blocktime, but in that case youre better off using it directly like im doing right now.

PR #323, opened 2022-10-31
added metrics for bonds #ELDER-187

SissonJ:
# Description

## Notable changes

- Added new contract query

## Next steps


wahlbergkyle (approved):

FloppyDisck (changes requested):
Would bump the count to `u32` just in case

FloppyDisck (approved):

PR #329, opened 2022-11-15
update snip20

GumboLimbo:
# Description
Makes snip20 impl more compliant with official standard

## Notable changes
Remove 'transfer-enabled' system

### DIFF
--- a/contracts/snip20/src/handle/transfers.rs
+++ b/contracts/snip20/src/handle/transfers.rs
@@ -14,8 +14,7 @@ use shade_protocol::{
     },
     contract_interfaces::snip20::{
         batch,
-        errors::transfer_disabled,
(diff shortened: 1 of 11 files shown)


FloppyDisck:
When you query this smart contract expecting the official SNIP20 config, it wont break it, therefore its technically SNIP20 compliant. 

These flags are important for making our SNIP20 easier to modify for staking tokens

GumboLimbo:
> When you query this smart contract expecting the official SNIP20 config, it wont break it, therefore its technically SNIP20 compliant.
> 
> These flags are important for making our SNIP20 easier to modify for staking tokens

Understood, the change to init config as well as the lack of automatically adding minters made this snip20 incompatible with my project. Imo it is fine if we leave this PR unmerged, these changes are only temporarily necessary, although there is potential for them to come up again

PR #336, opened 2023-03-14
Burn and mint

SissonJ:
Contract to facilitate the migration of Snip20s to the Snip25 standard 
Receives a token to burn and mints the new Snip25, sending the minted tokens to the user


contracts/snip20_migration/src/contract.rs
```diff
@@ -0,0 +1,146 @@
...
+        QueryMsg::Config {} => to_binary(&QueryAnswer::Config {
+            config: Config::load(deps.storage)?,
+        }),
+        QueryMsg::Metrics { token } => to_binary(&QueryAnswer::Metrics {
+            amount_minted: match AmountMinted::may_load(deps.storage, token.to_string())? {
+                Some(minted_amount) => Some(minted_amount.0),
+                None => None,
+            },
+        }),
+        QueryMsg::RegistragionStatus { token } => to_binary(&QueryAnswer::RegistrationStatus {
```
DrPresident, line 103: typo

contracts/snip20_migration/src/contract.rs
```diff
@@ -0,0 +1,146 @@
...
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::Config {} => to_binary(&QueryAnswer::Config {
+            config: Config::load(deps.storage)?,
+        }),
+        QueryMsg::Metrics { token } => to_binary(&QueryAnswer::Metrics {
+            amount_minted: match AmountMinted::may_load(deps.storage, token.to_string())? {
+                Some(minted_amount) => Some(minted_amount.0),
+                None => None,
+            },
```
DrPresident, line 101: Should error for invalid tokens, and return 0 for valid tokens with nothing minted not null

contracts/snip20_migration/src/contract.rs
```diff
@@ -0,0 +1,146 @@
...
+        registered_token.mint_token.clone().address.to_string(),
+    )?;
+
+    Ok(Response::default().add_message(mint_msg(
+        from.clone(),
+        burn_amount,
+        None,
+        None,
+        &registered_token.mint_token,
+    )?))
```
DrPresident, line 133: Tokens are never burned

contracts/snip20_migration/tests/mod.rs
DrPresident, line 1: This isn't a package, the name should be informational not `mod.rs` as that implies it contains package exports

contracts/snip20_migration/tests/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        snip20_migration::QueryAnswer::Config { config } => {
+            let expected_config = snip20_migration::Config {
+                admin: admin_auth.clone().into(),
+            };
+            if config == expected_config {
+                assert!(true);
+            } else {
+                assert!(false);
+            }
```
DrPresident, line 111: these asserts will fail and give no information on the failure. it should be `assert_eq!(config. expected_config, "helpful error info")`

contracts/snip20_migration/tests/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+        _ => assert!(false),
```
DrPresident, line 113: also an unhelpful failure, should be `panic!("helpful information")`

packages/shade_protocol/src/contract_interfaces/snip20_migration/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+#[cw_serde]
+pub struct RegisteredToken {
+    pub burn_token: Contract,
+    pub mint_token: Contract,
+}
+
+impl MapStorage<'static, String> for RegisteredToken {
+    const MAP: Map<'static, String, Self> = Map::new("registered_tokens-");
```
DrPresident, line 27: Storage stuff doesn't belong in this file

DrPresident (changes requested):

contracts/snip20_migration/src/contract.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+#[shd_entry_point]
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::Config {} => to_binary(&QueryAnswer::Config {
+            config: Config::load(deps.storage)?,
+        }),
+        QueryMsg::Metrics { token } => to_binary(&QueryAnswer::Metrics {
```
FloppyDisck, line 97: Note that this exposes the possibility of knowing user migrated balances

contracts/snip20_migration/src/contract.rs
FloppyDisck: Agree, replace option with a result

contracts/snip20_migration/tests/mod.rs
FloppyDisck: Agree, maybe this could be better off inside `src`

A separate test folder works when you have multiple files that share some helper functions

contracts/snip20_migration/tests/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    match (snip20_migration::QueryMsg::Metrics {
+        token: "lala".into(),
+    }
+    .test_query(&migration_contract, &mut chain)
+    .unwrap())
+    {
+        snip20_migration::QueryAnswer::Metrics { amount_minted } => match amount_minted {
+            None => assert!(true),
+            _ => assert!(false),
+        },
```
FloppyDisck, line 125:
```suggestion
        snip20_migration::QueryAnswer::Metrics { amount_minted } => assert!(amount_minted.is_none()),
```

FloppyDisck: many of the matches can be unwrapped into smaller things like these

packages/shade_protocol/src/contract_interfaces/snip20_migration/mod.rs
FloppyDisck: agree

FloppyDisck (changes requested):

contracts/snip20_migration/src/contract.rs
SissonJ: Tokens are burned by being received by the contract and having no way to leave, we can't guarantee that every snip20 we might want to use this contract for will have burn enabled

FloppyDisck: You could have a flag for it, or idk if the snip20 config shows that info but you could also query the config and store whether the snip has burn enabled or not

DrPresident: The only use-case currently for this contract is SHD, which can be burned. IMO we should actually burn those tokens so the total supply reflects the transition to the new token. The best solution here is what guy suggested and query the token config when it is registered and burn if able.

DrPresident (approved):

PR #343, opened 2023-05-24
feat: migration hotfix

SissonJ:
# Description

## Notable changes

- Added new contract query

## Next steps


DrPresident (commented):
We should add a test for the admin checks before redeploying

PR #358, opened 2023-08-11
Added Libraries

itsHaseebSaeed:
Libraries needed for liquidity book are added


contracts/liquidity_book/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[workspace]
+members = [
+    "libraries",
```
kent-3, line 3: This Cargo.toml file is defining a [virtual workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace). The members should also include the LB contracts.

contracts/liquidity_book/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = []
+# for quicker tests, cargo test --lib
+# for more explicit tests, cargo test --features=backtraces
+backtraces = ["shade-protocol/backtraces"]
+
+[dependencies]
```
kent-3, line 26: This table should be `[workspace.dependencies]` which will make them usable by all members of the workspace. [(more info)](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table)

This ensures all members will inherit the same versions of the dependencies. Here is an [example](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace) for how to do that.

contracts/liquidity_book/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+shade-protocol = { version = "0.1.0", path = "../../../packages/shade_protocol", features = [
+  "storage_plus",
+] }
+
+[dev-dependencies]
```
kent-3, line 31: In `[workspace.dependencies]` there is no separate table for dev-dependencies. You would lump them together here, then import them as dev-dependencies in the individual packages.

contracts/liquidity_book/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+debug = false
+rpath = false
+lto = true
+debug-assertions = false
+codegen-units = 1
+panic = 'abort'
+incremental = false
+overflow-checks = true
+
+[lib]
```
kent-3, line 17: I don't think you can use `[lib]` or `[features]` in a virtual workspace. These go in each package's manifest.

packages/shade_protocol/Cargo.toml
```diff
@@ -112,6 +112,7 @@ cosmwasm-std = { package = "secret-cosmwasm-std", version = "1.0.0" }
 cosmwasm-storage = { package = "secret-cosmwasm-storage", version = "1.0.0" }
 cosmwasm-schema = "1.1.5"
 contract-derive = { version = "0.1.0", path = "../contract_derive" }
+ethnum = { version = "1" }
```
kent-3, line 115: I'm not sure we should mess with the `shade_protocol` package. My thought is to have a separate folder like `/shade/packages/liquidity_orderbook/` for these libraries.

packages/shade_protocol/src/utils/mod.rs
```diff
@@ -23,6 +23,8 @@ pub mod generic_response;
 
 pub mod storage;
 
+pub mod liquidity_book;
```
kent-3, line 26: This feels wrong to me :skull:. I think we shouldn't mix with the core `shade_protocol` package.

kent-3 (changes requested):
Thanks for starting on this! For the first PR,  I think we should be extra careful not to mess up the existing cargo workspace and packages. Please see my other comments for my ideas on that.

contracts/liquidity_book/Cargo.toml
itsHaseebSaeed: My bad I forgot to remove this

kent-3: But I think we should keep it. Let's see how having a nested workspace goes.

itsHaseebSaeed: Nested workspaces don't work according to my knowledge

packages/shade_protocol/src/utils/mod.rs
itsHaseebSaeed: Gottca

contracts/liquidity_book/Cargo.toml
itsHaseebSaeed: I totally forgot to remove that 💀

contracts/liquidity_book/Cargo.toml
itsHaseebSaeed: Same as above.

contracts/liquidity_book/Cargo.toml
itsHaseebSaeed: removed this cargo.toml file for now.

packages/shade_protocol/Cargo.toml
itsHaseebSaeed: removed this cargo.toml file for now.

itsHaseebSaeed:
Created a separate package foe liquidity book

packages/shade_protocol/Cargo.toml
```diff
@@ -113,6 +113,7 @@ cosmwasm-storage = { package = "secret-cosmwasm-storage", version = "1.0.0" }
 cosmwasm-schema = "1.1.5"
 contract-derive = { version = "0.1.0", path = "../contract_derive" }
 
+
```
AustinWoetzel, line 116: lets remove this added line so as to not confuse the commit history on the main cargo.toml

itsHaseebSaeed: Fixed

AustinWoetzel:
are these going to be used by multiple contracts? I think you guys mentioned that liquidity book contained 5 different contracts. I would expect each of these contracts to get their own sub directory, then any helper functions like this would be imported from another directory elsewhere? @DrPresident what are your thoughts here?

image

PR #359, opened 2023-08-18
Refactored lb_factory

itsHaseebSaeed:
Added all the relevant contracts but lb_factory is refactored with _storage_plus_ (previously using secret toolkit). Other contracts are yet to be refactored.


packages/shade_protocol/Cargo.toml
```diff
@@ -112,12 +115,13 @@ cosmwasm-std = { package = "secret-cosmwasm-std", version = "1.0.0" }
...
 cosmwasm-schema = "1.1.5"
 contract-derive = { version = "0.1.0", path = "../contract_derive" }
+ethnum = { version = "1" }
 
 schemars = "0.8.9"
 serde = { version = "1.0.103", default-features = false, features = ["derive", "alloc"] }
 thiserror = "1.0"
 
-secret-storage-plus = { git = "https://github.com/securesecrets/secret-plus-utils", version = "0.13.4", optional = true }
+secret-storage-plus = { git = "https://github.com/securesecrets/secret-plus-utils", version = "0.13.4", optional = true, features = ["iterator"] }
```
itsHaseebSaeed, line 124: Added iterator feature

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+/// # Returns
+///
+/// * `bin_step_with_preset` - The list of bin steps.
+fn query_all_bin_steps(deps: Deps) -> Result<Binary> {
+    // NOTE: iterating over the keys of the PRESETS Keymap will return all available bin_steps
+    // not too confident with this implementation...
+
+    let mut bin_step_with_preset = Vec::<u16>::new();
+
+    let iterator = PRESETS
```
itsHaseebSaeed, line 1092: @DrPresident pls check this method's validity to iter over maps

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+    // TODO: decide on response output and format
+    Ok(Response::default())
+}
+
+/////////////// EXECUTE ///////////////
+
+#[entry_point]
+pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> Result<Response> {
+    match msg {
+        ExecuteMsg::SetLBPairImplementation {
```
DrPresident, line 70: Is this the factorys reference to the code id of the pair? if so this could be something like `set_lb_pair_code_id`, feels more clear & brief

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+            active_id,
+            bin_step,
+        } => try_create_lb_pair(deps, env, info, token_x, token_y, active_id, bin_step),
+        ExecuteMsg::SetLBPairIgnored {
+            token_x,
+            token_y,
+            bin_step,
+            ignored,
+        } => try_set_lb_pair_ignored(deps, env, info, token_x, token_y, bin_step, ignored),
+        ExecuteMsg::SetPreset {
```
DrPresident, line 88: Could this be `SetPairPreset`? (assuming thats the mechanism)

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+    token_x: TokenType,
+    token_y: TokenType,
+    active_id: u32,
+    bin_step: u16,
+) -> Result<Response> {
+    let state = CONFIG.load(deps.storage)?;
+
+    if !PRESETS.has(deps.storage, bin_step) {
+        return Err(Error::BinStepHasNoPreset { bin_step });
+    }
```
DrPresident, line 247: Can we be sure that cases like this never occur? like require a preset at init maybe, ideally if contract init is successful there should be no other steps required to start deploying pairs (or other basic mechanics)

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+    let is_owner = info.sender == state.owner;
+
+    if !_is_preset_open(preset.0 .0) && !is_owner {
+        return Err(Error::PresetIsLockedForUsers {
+            user: info.sender,
+            bin_step,
+        });
+    }
+
+    if !QUOTE_ASSET_WHITELIST
```
DrPresident, line 261: IIRC the "quote asset" term is for pairs like `quote / base`, assuming this asset references either side of the pair could it just be called `asset_whitelist`?

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+/// Sets whether the pair is ignored or not for routing, it will make the pair unusable by the router.
+///
+/// # Arguments
+///
+/// * `token_x` - The address of the first token of the pair.
+/// * `token_y` - The address of the second token of the pair.
+/// * `bin_step` - The bin step in basis point of the pair.
+/// * `ignored` - Whether to ignore (true) or not (false) the pair for routing.
+fn try_set_lb_pair_ignored(
```
DrPresident, line 368: Do we need this? I don't think the router should reference the factory except for possible the list of pairs, if it needs a blacklist of some sort I think it can have its own

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+    Ok(Response::default().add_attribute_plaintext("set preset", bin_step.to_string()))
+}
+
+/// Sets if the preset is open or not to be used by users
+///
+/// # Arguments
+///
+/// * `bin_step` - The bin step in basis point, used to calculate the price
+/// * `is_open` - Whether the preset is open or not
+fn try_set_preset_open_state(
```
DrPresident, line 501: I'm confused on what this means, for user created pairs the preset wouldn't be used at all if its not set to "open"?

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -0,0 +1,1261 @@
...
+/// * `token_y` - The address of the second token
+/// * `bin_step` - The bin step in basis point, used to calculate the price
+/// * `base_factor` - The base factor, used to calculate the base fee, baseFee = baseFactor * binStep
+/// * `filter_period` - The period where the accumulator value is untouched, prevent spam
+/// * `decay_period` - The period where the accumulator value is decayed, by the reduction factor
+/// * `reduction_factor` - The reduction factor, used to calculate the reduction of the accumulator
+/// * `variable_fee_control` - The variable fee control, used to control the variable fee, can be 0 to disable it
+/// * `protocol_share` - The share of the fees received by the protocol
+/// * `max_volatility_accumulator` - The max value of volatility accumulator
+fn try_set_fee_parameters_on_pair(
```
DrPresident, line 570: I think this can just be done on the pair contract, unless the factory does mass applications that might be nice, but then wouldn't the factory need admin on the pair?

contracts/liquidity_book/lb_factory/src/msg.rs
```diff
@@ -0,0 +1,230 @@
...
+        // u24
+        max_volatility_accumulator: u32,
+    },
+    SetFeeRecipient {
+        fee_recipient: Addr,
+    },
+    SetFlashLoanFee {
+        flash_loan_fee: u8,
+    },
+    AddQuoteAsset {
```
DrPresident, line 81: Could this be `whitelist_asset`?

contracts/liquidity_book/lb_factory/src/msg.rs
```diff
@@ -0,0 +1,230 @@
...
+        asset: TokenType,
+    },
+    RemoveQuoteAsset {
+        asset: TokenType,
+    },
+    ForceDecay {
```
DrPresident, line 87: What exactly is `decay`?

contracts/liquidity_book/lb_quoter/src/msg.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub factory: Addr,
+}
+
+#[cw_serde]
+pub enum ExecuteMsg {
+    Swap { swap_for_y: bool, to: Addr },
+}
```
DrPresident, line 12: what exactly does this contract do?

contracts/liquidity_book/lb_router/src/msg.rs
```diff
@@ -0,0 +1,132 @@
...
+        token_y: TokenType,
+        active_id: u32,
+        bin_step: u16,
+    },
+    Receive {
+        from: String,
+        msg: Option<Binary>,
+        amount: Uint128,
+    },
+    SwapTokensForExact {
```
DrPresident, line 25: Can swapping just be done on `Receive`? or does this function do anything unique?

contracts/liquidity_book/lb_router/src/msg.rs
```diff
@@ -0,0 +1,132 @@
...
+        /// The token type to swap from.
+        offer: TokenAmount,
+        expected_return: Option<Uint128>,
+        path: Vec<Hop>,
+        recipient: Option<String>,
+    },
+    RegisterSNIP20Token {
```
DrPresident, line 32: Names like `SNIP20` get serialized as `s_n_i_p20`, I would just go with `register_token`.
Then there could also be another field like `register_multi_token` for the snip1155?

contracts/liquidity_book/lb_router/src/msg.rs
```diff
@@ -0,0 +1,132 @@
...
+        token_addr: String,
+        token_code_hash: String,
+    },
+    RecoverFunds {
```
DrPresident, line 36: Make sure this works for both Layer 1 & snip20 tokens

contracts/liquidity_book/lb_router/src/msg.rs
```diff
@@ -0,0 +1,132 @@
...
+    GetFactory {},
+
+    #[returns(IdFromPriceResponse)]
+    GetIdFromPrice {
+        lb_pair: ContractInfo,
+        price: Uint256,
+    },
+
+    #[returns(PriceFromIdResponse)]
+    GetPriceFromId { lb_pair: ContractInfo, id: u32 },
```
DrPresident, line 77: Confused on what this query does

DrPresident (approved):
I think in instead of the `ilbfactory` style I'd prefer doing `lb_factory` for the `contract_interfaces`

The structure looks good! I had a few things in there and some questions that we can discuss in the meeting.

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: well bin step can range from [0.0001, 0.01]. so we can't have preset for all the binsteps. Ideally we would set preset for those binsteps and then try to create a pair. or what we can do it that if the presets doesn't exist we just used if any preset sent.

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: Well the idea behind is that you won't be able to create a pool with non-whitelisted asset. Reason is to keep the value of Quote asset and to avoid any error like BTC should only be Quote asset and base asset would be USDT.

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: Agree

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: Well it means that we can set which presets can be used by users(nor admins) to createLbPool

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: I agree, let's comment it for now

contracts/liquidity_book/lb_factory/src/msg.rs
itsHaseebSaeed: So in liquidity book volatility is measured to calculate the dynamic fee. Decay is the deadline after which the momentum of the volatility is lost and the volatility is set to zero

contracts/liquidity_book/lb_quoter/src/msg.rs
itsHaseebSaeed: It Helps  to determine best path through multiple markets. But I don't think we'll use it. We'll remove after the structure if cleared.

contracts/liquidity_book/lb_router/src/msg.rs
itsHaseebSaeed: It's was half done when deadline came need to fix this according your requirements. I wouldn't look at this ngl. 💀

contracts/liquidity_book/lb_router/src/msg.rs
itsHaseebSaeed: Helps get price from based on bin id. I don't think we need that since we can simply query lb-pair. P.S: Bin Prices are fixed at the start.

PR #364, opened 2023-10-03
Added liquidity_book multi_tests

itsHaseebSaeed:


contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -556,70 +559,74 @@ fn try_remove_preset(
...
+/// * `token_y` - The address of the second token
+/// * `bin_step` - The bin step in basis point, used to calculate the price
+/// * `base_factor` - The base factor, used to calculate the base fee, baseFee = baseFactor * binStep
+/// * `filter_period` - The period where the accumulator value is untouched, prevent spam
+/// * `decay_period` - The period where the accumulator value is decayed, by the reduction factor
+/// * `reduction_factor` - The reduction factor, used to calculate the reduction of the accumulator
+/// * `variable_fee_control` - The variable fee control, used to control the variable fee, can be 0 to disable it
+/// * `protocol_share` - The share of the fees received by the protocol
+/// * `max_volatility_accumulator` - The max value of volatility accumulator
+fn try_set_fee_parameters_on_pair(
```
DrPresident, line 576: Is this like a global fee configuration for all pairs? if its for a specific pair we should probably have it on the pair itself. 
If it is global, can we just call it like `update_default_fees`

contracts/liquidity_book/lb_factory/src/contract.rs
```diff
@@ -556,70 +559,74 @@ fn try_remove_preset(
...
+            token_y: token_b.unique_key(),
+            bin_step,
+        })?
+        .lb_pair;
+
+    let mut response = Response::new();
+
+    response = response.add_message(CosmosMsg::Wasm(WasmMsg::Execute {
+        contract_addr: lb_pair.contract.address.to_string(),
+        code_hash: lb_pair.contract.code_hash,
```
DrPresident, line 615: I'm personally a fan of using `to_cosmos_msg` for things like this, we use it in alot of our helper functions. You just need to add like a trait to the `ExecuteMsg` struct and then have that trait imported wherever you intend to use it

contracts/liquidity_book/lb_pair/src/contract.rs
```diff
@@ -183,15 +188,45 @@ pub fn instantiate(
...
 #[shd_entry_point]
 pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> Result<Response> {
     match msg {
-        ExecuteMsg::Swap {
-            swap_for_y,
+        ExecuteMsg::Receive(msg) => {
+            let checked_addr = deps.api.addr_validate(&msg.from)?;
+            receiver_callback(deps, env, info, checked_addr, msg.amount, msg.msg)
+        }
+        ExecuteMsg::SwapTokens {
```
DrPresident, line 199: I don't think we need Layer-1 support for this contract, especially if we don't have full support for creating a pair with layer-1's and interacting with them for all mechanisms (fees etc.)

DrPresident (approved):
For the `shadeswap_shared` package, do you need anything from that package except for like `ExecuteMsg` & `QueryMsg` from some of the contracts? it seems it may be possible to easily pull in just the desired structs. Not a big deal but would be nice if there's a clear set of items needed vs all the extra logic included in that package.

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: no it's for specific pairs. Not global

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: Replaced!!!

contracts/liquidity_book/lb_pair/src/contract.rs
itsHaseebSaeed: Yeap it's just one nice thing to have. Already spend significant time on this 💀

contracts/liquidity_book/lb_factory/src/contract.rs
itsHaseebSaeed: @DrPresident do check the PR.

PR #390, opened 2024-01-10
Milestone 3 completion

itsHaseebSaeed:
- Added Query_auth 

- Updated lb_staking staking algorithm 

- Added Protections to disallow recovering the staked token


contracts/liquidity_book/lb_staking/src/execute.rs
```diff
@@ -125,14 +158,60 @@ pub fn try_stake(
...
+    let token_id: u32 = token_id_string.parse().unwrap();
+
     //LOADING general use readonly stores
-    let epoch_obj = match EPOCH_STORE.may_load(deps.storage, state.epoch_index)? {
+    let mut epoch_obj = match EPOCH_STORE.may_load(deps.storage, state.epoch_index)? {
         Some(e) => Ok(e),
         None => Err(StdError::generic_err("Reward token storage already exists")),
     }?;
+
+    loop {
```
DrPresident: I don't think this loop is needed, if it is it should be reworked to a for loop or iterating an object/series

contracts/liquidity_book/lb_staking/src/execute.rs
```diff
@@ -257,12 +332,61 @@ pub fn try_unstake(
     }
 
     let mut state = STATE.load(deps.storage)?;
-    let epoch_obj = EPOCH_STORE
+    let mut epoch_obj = EPOCH_STORE
         .may_load(deps.storage, state.epoch_index)?
         .ok_or_else(|| StdError::generic_err("Reward token storage does not exist"))?;
 
+    loop {
```
DrPresident: Same here

